PR #860, opened 2020-03-23
Enhance Dialog

### LINKED ISSUE #617: useTransition re-renders / mounts twice
I use `useTransition` for a transition between screens.
I've implemented this based on the Simple Transition example.

Unfortunately, the entering component renders twice which causes a flickering in my special case (scroll down using ref). I've extended the example for presentation purposes. Please see the `console.log`. https://codesandbox.io/s/y7wwzyvjkz?fontsize=14

Is there a simple way to avoid this behaviour or is it a bug?

Thanks!

robertu7:
* Use `useSpring` instead of `useTransition` to fix [mount twice](https://github.com/react-spring/react-spring/issues/617)
* Add `slideIn` prop, and apply only in a small number of dialogs, like dropdowns [1]
* [Focus editor](https://github.com/thematters/matters-web/pull/860/files#diff-4fb631e33b503bd009a4a1498e4da85eR37) when the animation gets rest
* Support hit the escape key to close the dialog
* Bump `next` to `9.3.4`

[1] The slide in animation on mobile is still buggy. Not sure if it's caused by wrong calculation of `translateY` in percentage unit since mostly appear in dynamic height dialogs (forms, editor, fingerprint, etc.).


guoliu (approved):
Nice find on useTransition re-render issue!

PR #1255, opened 2020-06-29
English version of texts

proformatters:
Hi guys, here is the new English version of texts. I found that there are some repetitive texts, and according to my memory, there should be many other texts that have not been included. We need to continuously update.


robertu7:
closed by #1822

Issue #1437, opened 2020-08-12
Storybook for component library


guoliu:
**Is your feature request related to a problem? Please describe.**

As our component library and platform functionality grows, we need an easier way to visualize our components for both designers and developers.

**Describe the solution you'd like**

We can setup [Storybook](https://storybook.js.org/) for frontend repo. We can start with running Storybook locally, then push the static version it to s3 and serve it via matters url if needed.

Ginhing:
Hi, just mention it the PR is ready. Feel free to let me know if there is any problems 😀

guoliu:
@Ginhing the PR looks good! Are you able to merge it in?

Ginhing:
Oh, I assume owner or maintainer will merge it and I have no permission to merge at the same time.

devformatters:
Nice work. Close it. 💪🏻

PR #1483, opened 2020-09-05
docs(storybook): Introduce Storybook

Ginhing:
Since nextjs has done a great job on bundling but storybook can't enjoy the results. We need to add
custom configuration in `.storybook/main.js` once again.

re #1437


guoliu (approved):
Looks good! Let's get it in 🌵

Ginhing:
messed up, see #1564 instead.

PR #1576, opened 2020-10-13
feat(tag): revise description in user's tag feed due to now listing tags owned and maintained

devformatters:
As title.


robertu7 (approved):

guoliu (approved):
🥕

PR #1581, opened 2020-10-19
Fix Editor UI

robertu7:
* Fix the wrong position of placehoder;
* Using `height` instead of `max-height` to increase default editor height for larger clickable hot area;
* Fix extra container scrollbar by hiding `Layout > aside` on mobile;


devformatters (approved):
Go. 🚰

PR #1585, opened 2020-10-19
feat(verification): use verification link in password resetting flow

robertu7:


src/components/Forms/SignUpForm/Init.tsx
```diff
@@ -120,31 +115,14 @@ const Init: React.FC<FormProps> = ({
...
+        setFieldError('email', messages[codes[0]])
 
         if (refreshToken) {
           refreshToken()
         }
-
-        setSubmitting(false)
       }
+
+      setSubmitting(false)
```
devformatters, line 125: There's a `submitCallback` before `setSubmitting`, so it might cause state update for a unmounted component. 🤔

devformatters (approved):
Go. 🐚

src/components/Forms/SignUpForm/Init.tsx
robertu7: Nice catch! Fix all other forms too.

PR #1589, opened 2020-10-21
fix(editor): revert #1567 and make figure non-full width

robertu7:
Previous bugfix #1567 broke the behavior of the editor toolbar, it won't appear when we focus on newline.


devformatters (approved):
Nice fixes. 🧿

PR #1652, opened 2020-11-25
fix(auth token): retrieve and use auth token from localStorage only during static build

guoliu:
As title.

Also added `node-fetch` to devDependencies, since it's used in `bin/buildFragmentTypes.js`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -25282,6 +25282,12 @@
         }
       }
     },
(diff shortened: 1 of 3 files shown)


robertu7 (approved):

devformatters (approved):
Go. 🚛

PR #1674, opened 2020-12-07
fix(onboarding-tasks): store init state separately based on viewer id

robertu7:
* Remove clear state on logout since it's stored separately
* Fix `setPrivateFetched`


devformatters (approved):
Nice fix. 🙌

PR #1675, opened 2020-12-07
Fix <LikeCoinDialog> auto closed on Liker ID finish setup

robertu7:
This issue was caused by `viewer.likerId`'s changing, which will trigger UI update and make dialog auto closes.


guoliu (approved):
🌾

PR #1694, opened 2020-12-18
fix(form): rollback formik to 2.1.x

robertu7:
`formik@2.2.x` causes donation input validation incorrectly.


devformatters (approved):
Go. 🧑🏻‍💼

PR #1714, opened 2020-12-29
[hotfix - master] fix(style): add safe-area to editor toolbar

robertu7:


robertu7:
https://github.com/thematters/matters-web/pull/1716

PR #1721, opened 2021-01-04
fix(style): add spacer above the tabs of follow page

robertu7:
Without a spacer, the bottom borders of header and tabs are hidden;

image


devformatters (approved):
Go. 🦸🏼‍♂️

guoliu (approved):

PR #1732, opened 2021-01-07
feat(circle): edit circle profile page

robertu7:


src/common/utils/route.ts
```diff
@@ -95,6 +95,16 @@ export const toPath = (args: ToPathArgs): { href: string } => {
...
       }
     }
+    case 'circleSettings': {
+      return {
+        href: `/~${args.circle.name}/settings`,
+      }
+    }
+    case 'circleEditProfile': {
+      return {
+        href: `/~${args.circle.name}/settings/edit-profile`,
```
devformatters, line 105: why not `/profile/edit` ?

devformatters (approved):
Let's try it. 🎓

src/common/utils/route.ts
robertu7: You mean `/settings/profile/edit` or just `/profile/edit`?

PR #1764, opened 2021-01-15
fix(editor): incorrect editing distance during revision

devformatters:
Note: CI checker will pass until new version matters-editor published.


guoliu (approved):
🌿

src/views/ArticleDetail/EditMode/index.tsx
```diff
@@ -42,7 +42,7 @@ const EditMode: React.FC<EditModeProps> = ({ article, onCancel, onSaved }) => {
   const isLargeUp = useResponsive('lg-up')
 
   // staging editing data
-  const [content, editContent] = useState<string | null>(null)
+  const [editData, setEditData] = useState<Record<string, any>>({})
```
robertu7, line 45: export type from editor package and use it here?

robertu7 (approved):

src/views/ArticleDetail/EditMode/index.tsx
devformatters: Let's put it in the following PRs.

PR #1767, opened 2021-01-19
feat(circle): add members and followers page for Circle

devformatters:


src/pages/[name]/followers.tsx
```diff
@@ -1,3 +1,30 @@
...
+
+import { getNameType } from '~/common/utils'
+
+const Followers = () => {
+  const router = useRouter()
+  const nameType = getNameType({ router })
+
+  switch (nameType) {
+    case 'circle': {
+      return <CircleFollowers />
```
Comment (robertu7) [inline src/pages/[name]/followers.tsx:15]: Like `[name]/index.tsx`, put them into dynamic import with `ssr: true`?

robertu7 (approved):
🏔️

src/pages/[name]/followers.tsx
Comment (devformatters) [reply src/pages/[name]/followers.tsx]: Nice point. Revised.

PR #1775, opened 2021-01-21
refactor(layout): add `<Layout.FixedMain>`

robertu7:
* add `<Layout.FixedMain>` as a fixed container for `<Toast>` and `<NavBar>` etc.; 
* transforms anonymous components into named components[ to make sure they work with Fast Refresh](https://nextjs.org/docs/advanced-features/codemods#name-default-component);


devformatters (approved):
Nice hard work. 🦸🏼‍♂️

PR #1783, opened 2021-01-28
feat(feature-flag): add feature flag for Circle

devformatters:


src/views/ArticleDetail/EditMode/index.tsx
```diff
@@ -166,7 +168,7 @@ const EditMode: React.FC<EditModeProps> = ({ article, onCancel, onSaved }) => {
             disabled={isEditDisabled}
           />
 
-          {toggleCircle && (
+          {toggleCircle && features.circle_management && (
```
robertu7, line 171: Forgot adding  `features.circle_management` at `<BottomBar>` L236?

src/components/Hook/useFeatures.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * This hook is a wrapper of accessing feature flag context.
+ *
+ * Usage:
+ *
+ * ```jsx
+ *   const features = useFeatures()
+ * ```
+ */
+export const useFeatures = () => {
```
robertu7, line 14: Nice abstraction, would it be good if applying to the rest of the places, don't leave legacy :)

robertu7 (changes requested):

src/views/ArticleDetail/EditMode/index.tsx
devformatters: Totally missed it. Now add it.

src/components/Hook/useFeatures.ts
devformatters: Done.

robertu7 (approved):

PR #1777, opened 2021-01-21
Circle Subscription

robertu7:
* Add subscribe and unsubscribe circle dialog;
* Clean code with adding `useRoute` hook and applying `useDialogSwitch`;

PR #1786, opened 2021-02-01
refactor(error): ignore default error catching for form mutations

robertu7:
During the implementation of #1777, I found some errors like `CIRCLE_NOT_FOUND` was caught and then shown and skip the error throwing, which makes mutation inside [`try...catch...`](https://github.com/thematters/matters-web/pull/1777/files#diff-a9913444763a337cf658f9b2fbe0c7e4b2e1e67614efa87e97ed1027cfe0c5adR71-R82) doesn't work as we expected [1].  

In these changes, forms ignore the default error catching behavior and show the error message in the form field instead [2].

[1] Showing toast and subscription becomes complete
image

[2] 
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14922,17 +14922,29 @@
       "dev": true
     },
     "formik": {
-      "version": "2.2.6",
-      "resolved": "https://registry.npmjs.org/formik/-/formik-2.2.6.tgz",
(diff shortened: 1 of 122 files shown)


guoliu (approved):
Nice! 🗿

PR #1830, opened 2021-03-01
User dialogs instead of user pages

robertu7:
* Use user dialogs to replace user list pages for better UX;
* Fix circle pinned broadcast `text-indent`;
* Fix spacing of `<CircleBanner>`;


jeremyOk (approved):

devformatters (approved):
Go. 🕵🏾

PR #1839, opened 2021-03-02
fix(circle): fix incorrect rendering on circle detail pages

robertu7:
* fix incorrect rendering on circle detail pages;
* wrap comment list with `<InfinteScroll>`;


devformatters (approved):
Go. 🕹

PR #1871, opened 2021-03-08
fix(circle): add updater for follow or unfollow

devformatters:


guoliu (approved):

robertu7 (approved):
though lots of schema dependencies, let's try!

PR #1907, opened 2021-03-15
ci: prevent deployment on pull request

robertu7:
* prevent triggering `deploy.yml` workflow on pull request;
* trigger `storybook.yml` only if the pull request was labeled with `storybook` [1];

[1] Relabel is needed if the PR has more commits;


devformatters (approved):
Let's test it. 🎃

PR #1917, opened 2021-03-19
Revamp: About Us; Refactor layout;

robertu7:
* Revamp about us page;
* Refactor Grid System with CSS Grid;


devformatters (approved):
Nice refactor. Let's try it. 👔

PR #1957, opened 2021-04-07
chore(deps): update deps

robertu7:
* Remove `routerPush` since it already been [fixed in Next.js 10.1](https://nextjs.org/blog/next-10-1#router-methods-scroll-to-top).


devformatters (approved):
Nice upgrade. 🫀

PR #1968, opened 2021-04-12
Redirection on article with old hashes;

robertu7:
* redirect article to latest hash URL;
* correct summary on draft editor;


src/views/ArticleDetail/index.tsx
```diff
@@ -138,6 +138,33 @@ const ArticleDetail = () => {
     loadPrivate()
   }, [article?.mediaHash, viewer.id])
 
+  // redirect to latest published article
+  const latestArticle = article?.drafts?.filter(
```
guoliu, line 142: Nice! Putting the decision of newest version on ssr server is neat.

guoliu (approved):

PR #1979, opened 2021-04-15
fix(circle): fix missing save button on circle management dialog

robertu7:
Also,

* Revise fallback `og:description`;


devformatters (approved):
Nice. 🐔

PR #1986, opened 2021-04-16
fix(style): fix CSS Grid blowout

robertu7:
The behavior of CSS Grid on Firefox may break the layout when the content width overflow.

ref: https://stackoverflow.com/questions/43311943/prevent-content-from-expanding-grid-items


devformatters (approved):
Go. 🥻

PR #2027, opened 2021-05-06
feat(revision): show conformation popup before exit edit mode

robertu7:


devformatters (approved):
Let team review it. 👘

PR #2033, opened 2021-05-07
fix the wording about IPFS

catdingding:


guoliu (approved):
Looks good. There's are more wordings to be cleaned up, will follow up with another PR.

PR #2093, opened 2021-06-28
fix(router): fix back button behavior after publishing draft

robertu7:
**Before**
Redirect back to the draft detail page;

**Current**
Redirect back to the previous page of draft detail page;


devformatters (approved):
Go. 🌵

PR #2094, opened 2021-06-28
Upgrade to Next.js 11

robertu7:


src/components/index.tsx
```diff
@@ -24,7 +24,8 @@ export * from './Expandable'
 export * from './Slides'
 export * from './CopyToClipboard'
 export * from './Context'
-export * from './GQL'
+export * from './GQL/hooks'
```
robertu7, line 27: An error `usePublicQuery is not a function` was threw on dev & build, not sure if it's by Webpack 5.

devformatters (approved):
Go. 🧚🏼‍♂️

PR #2111, opened 2021-07-02
[hotfix - master] fix(static): fix static file serving

robertu7:
* Use Next.js built-in [Automatic Image Optimization](https://nextjs.org/docs/basic-features/image-optimization) instead of `next-optimized-images`;
* Rename `` to `<ResponsiveImage>` to distinguish with [`<Image>`](https://nextjs.org/docs/basic-features/image-optimization#image-component);

PR #2112, opened 2021-07-02
[hotfix - develop] fix(static): fix static file serving

robertu7:
[Changes](https://github.com/thematters/matters-web/pull/2111)


devformatters (approved):
Go. 🦾

PR #2144, opened 2021-07-23
Enhancement of following feed

robertu7:
* Break down following feed components and fragments;
* Fix missing activity dates;
* Add style on hover/focus head username;


guoliu (approved):
Let's try it 🌵

PR #2159, opened 2021-08-02
Fix accessibility bugs.

FledgeXu:
relative: #2158
This PR fixed the bug which is that VoiceOver can not read the Appreciation count and Donation count.


src/common/enums/text.ts
```diff
@@ -22,6 +22,8 @@ export const TEXT = {
     allTopics: '熱議廣場',
     appreciationsReceived: '讚賞我的',
     appreciationsSent: '我讚賞的',
+    appreciate: '讚賞作品',
```
robertu7, line 25: please also add `zh_hans` and `en` versions of translations. In this PR, I believe it's overkill to move the aria text to `enums/text.ts` since it's for reusable translations, also, we haven't added i18n support for aria.

src/views/ArticleDetail/AppreciationButton/AppreciateButton.tsx
```diff
@@ -38,7 +39,9 @@ const AppreciateButton: React.FC<AppreciateButtonProps> = ({
       <Button
         spacing={['xtight', 'xtight']}
         bgActiveColor="grey-lighter"
-        aria-label="讚賞作品"
+        aria-label={`${TEXT.zh_hant.appreciate} ${
```
robertu7, line 42: Could it add to inside [count text](https://github.com/thematters/matters-web/blob/7240933f082fc806ac64c60b7cea4da335901bba/src/views/ArticleDetail/AppreciationButton/AppreciateButton.tsx#L75) directly? "讚賞作品" is descriptive text of the action button, mixes with "讚賞數量: XXX" seems not intuitive enough?

src/views/ArticleDetail/Toolbar/DonationButton/index.tsx
```diff
@@ -78,7 +78,13 @@ const DonationButton = ({ article, disabled }: DonationButtonProps) => {
           <Button
             spacing={['xtight', 'xtight']}
             bgActiveColor="grey-lighter"
-            aria-label={TEXT.zh_hant.donation}
+            aria-label={`${TEXT.zh_hant.donation} ${
+              TEXT.zh_hant.donationCount
```
robertu7, line 82: Same as above.

robertu7 (changes requested):
Congrats on creating your first PR!

I have left some comments, please check. 

BTW, `package-lock.json` was changed, please make sure you are running the right versions of node (12.x) & npm (6.x) and reinstall packages (`npm i`).

src/common/enums/text.ts
FledgeXu: This can move back.

src/views/ArticleDetail/AppreciationButton/AppreciateButton.tsx
FledgeXu: `aria-label` actually is more like a description of what we can see of an element.  
In this case, We actually can see two different parts: `讚賞作品`  and the number of `讚賞作品`.
image
When we added it, it will look like this:
image

src/views/ArticleDetail/Toolbar/DonationButton/index.tsx
FledgeXu: Same as above.

robertu7 (approved):

PR #2160, opened 2021-08-02
StackedAreaChart

robertu7:
[Demo](https://thematters.github.io/matters-web/storybook/feature/chart/iframe.html?id=components-charts--all&args=&viewMode=story)


devformatters (approved):
Nice work. Looks good. 🤳🏻

PR #2185, opened 2021-08-17
fix(deps): fix es module error with downgrading d3; update deps;

robertu7:
https://stackoverflow.com/questions/68126052/how-do-i-resolve-err-require-esm-must-use-import-to-load-es-module-when-usi


devformatters (approved):
Have the same problem. Nice fix. 🧩

PR #2190, opened 2021-08-18
feat(analytics): Card exposure analytics

guoliu:
Add card exposure record for homes feeds and follow feed.

Record logic:
* When card enters view, start 500ms timer, and fire analytic event when timer is up
* When card leaves view, cancel timer
* Every card is only recorded once during one rendering

Event properties (mostly same with "click_feed" event):
* id: article id
* feedType: type of feed, can be "following", "hottest", "newest" or "icymi"
* contentType: type of card, can be "article", "UserDonateArticleActivity", "ArticleRecommendationActivity", "UserRecommendationActivity" and "UserPublishArticleActivity"
* location: location of card. When card is in a row, location use "." to separate column and rows. E.g., for card at row 3 column 2, location is "3.2"
* userId: user id, "null" for visitor


src/components/Analytics/CardExposureTracker/index.tsx
```diff
@@ -0,0 +1,46 @@
...
+              contentType,
+              location,
+              userId: viewer.id,
+              id,
+            })
+          }, 500)
+          setTimerId(timer)
+        }
+      }}
+      onLeave={() => window.clearTimeout(timerId)}
```
devformatters, line 43: Maybe put timer initialization in `useEffect` ?

guoliu: In this way timer will start when component is rendered; but the component might not be in view when rendered (e.g. when the feed is loaded but lower than the view), that's why I used waypoint to start the timer.

Is their any down side in initializing timer in with "onEnter"?

jeremyOk (approved):

src/components/Analytics/CardExposureTracker/index.tsx
devformatters: I see it. Just assume we can handle timer initialization and removal better in `useEffect`.

robertu7: By putting in `useEffect`, we can clear the timer on component unmount, more like a better practice. 🤔

guoliu: @devformatters @robertu7 I added the timer removal logic in `useEffect`. The setup is still in `onEnter`, since I couldn't find a way to set it up in `useEffect` but start later when component enter view 🤔

devformatters: How about:

```jsx

const [active, setActive] = useState(false)

useEffect(() => {

  if (active) {
    ...
  }
  
}, [])

<WayPoint
  onEnter={() =>  {
    ...
    if (active === false) {
      setActive(true)
    }
  }} 
  onLeave={() => {
     ...
  }}
 />

```

devformatters (approved):
Let's try it. 🪱

PR #2191, opened 2021-08-19
Enhance circle analytics

robertu7:
* rename income fields;
* add color indicator to tooltip and info tiles;


devformatters (approved):
Go. 🪖

PR #2228, opened 2021-08-30
fix(analytics): card exposure in sliders, page view on first load, id on `click_feed`

guoliu:
* Add 1s wait time for listener to be ready before firing page view and identify user 
* Use two Waypoint to measure card exposer in sliders, one for vertical one for horizontal
* Add `id` field on `click_feed` events


jeremyOk (approved):

robertu7 (approved):
Let's try.

devformatters (approved):
Go. 💨

Issue #2254, opened 2021-09-23
Old articles cannot render image


guoliu:
Examples: 
* https://matters.news/@935593258/%E6%96%87%E9%9D%A9%E5%AE%9E%E9%AA%8C-%E4%B8%BA%E4%BA%86%E5%85%A5%E6%88%8F-%E8%BF%99%E7%BE%A4%E6%BC%94%E5%91%98%E5%9B%9E%E5%88%B0%E4%BA%86%E7%96%AF%E7%8B%82%E7%9A%841969-zdpuArRLyD1g7voKCJo85beSDZ2JyM9REqKYQo2EwxB7wqQdd
* https://matters.news/@yinancl/hi-%E9%82%A3%E4%B8%AA%E6%B6%88%E5%A4%B1%E7%9A%84%E5%8F%97%E5%AE%B3%E8%80%85%E5%9B%9E%E6%9D%A5%E4%BA%86-zdpuAt5fS3H74h7YZeRUmxYM7DPwHCu4rUKDNsGa5u8YdMo9h

tx0c:
the first article has 5 imgs and have 5 times of this error in JS console:

```console
5 Refused to load the image '<URL>' because it violates the following Content Security Policy directive: "img-src 'self' data: assets.matters.news <URL>".
```

these old articles are referencing from `srcSet=https://matters-server-production.s3-ap-southeast-1.amazonaws.com/embed/...`

```
<meta http-equiv="Content-Security-Policy" content=" ...  img-src &#x27;self&#x27; data: assets.matters.news www.google-analytics.com ....
```

might just need an update of Content-Security-Policy

tx0c:
it's weird why do we need the google-analytics in the img-src ?

from https://github.com/thematters/matters-web/blame/fe36377cfc0dad880c66b9524bd195fbada34d4f/src/common/enums/csp.ts#L45

it was added in https://github.com/thematters/matters-web/pull/2049/files#diff-88097b50266c1bffddb81cb6020b72e2d87e2ab286dc3262dcdb29f312f59af5R43 by @robertu7   4 months ago

PR #2255, opened 2021-09-24
update Content-Security-Policy: add matters s3 url

tx0c:
should fix large part of #2254  as @robertu7 mentioned there were still others servers when users were copying pictures

these old articles were referencing from `srcSet=https://matters-server-production.s3-ap-southeast-1.amazonaws.com/embed/...` directly

### DIFF
--- a/src/common/enums/csp.ts
+++ b/src/common/enums/csp.ts
@@ -39,6 +39,9 @@ const IMG_SRC = [
   'data:',
   process.env.NEXT_PUBLIC_ASSET_DOMAIN,
 
+  // for some old articles were using this s3 urls directly
+  'matters-server-production.s3-ap-southeast-1.amazonaws.com',
+
   // GA
   'www.google-analytics.com',
 ].join(' ')


src/common/enums/csp.ts
```diff
@@ -39,6 +39,9 @@ const IMG_SRC = [
   'data:',
   process.env.NEXT_PUBLIC_ASSET_DOMAIN,
 
+  // for some old articles were using this s3 urls directly
+  'matters-server-production.s3-ap-southeast-1.amazonaws.com',
+
   // GA
   'www.google-analytics.com',
```
tx0c, line 46: why do we need google-analytics here?

robertu7: GA-related domains are based on https://developers.google.com/tag-manager/web/csp#universal_analytics_google_analytics , looks we still need it.

robertu7 (approved):
LGTM

src/common/enums/csp.ts
guoliu: Since we have switched to Firebase, we should also review if we still need GA (injected [here](https://github.com/thematters/matters-web/blob/develop/src/pages/_document.tsx#L47))

guoliu (approved):

tx0c:
running local dev with `NEXT_PUBLIC_API_URL=https://server.matters.news/graphql` can show http://localhost:3000/@935593258/文革实验-为了入戏-这群演员回到了疯狂的1969-zdpuArRLyD1g7voKCJo85beSDZ2JyM9REqKYQo2EwxB7wqQdd with all 5 pictures no problem:

image

Issue #2263, opened 2021-09-28
JSON-LD for article type


guoliu:
Add JSON-LD for article type to enhance search appearance on Google. 

Ref: https://developers.google.com/search/docs/advanced/structured-data/intro-structured-data

PR #2260, opened 2021-09-28
feat/seo-json-ld: enhance articles SEO with Json-LD data

tx0c:
this is so far adding Json-LD to articles page (`/@userName/...-mediaHash`) only

PR #2293, opened 2021-10-22
call serialize to replace </script> tag, avoid xss

tx0c:
https://redux.js.org/usage/server-rendering/#security-considerations

if we want to embed this, the minimum is
https://github.com/yahoo/serialize-javascript/blob/main/index.js#L19-L31

```js
// Mapping of unsafe HTML and invalid JavaScript line terminator chars to their
// Unicode char counterparts which are safe to use in JavaScript strings.
const ESCAPED_CHARS = {
    '<'     : '\\u003C',
    '>'     : '\\u003E',
    '/'     : '\\u002F',
    '\u2028': '\\u2028',
    '\u2029': '\\u2029'
};

// then...
JSON.stringify(data) .replace(/[<>\/\u2028\u2029]/g, unsafeChar => ESCAPED_CHARS[unsafeChar]);
```

1. JSON-LD for article type #2263
2. feat/seo-json-ld: enhance articles SEO with Json-LD data #2260


guoliu:
@tx0c shall we merge this in? This PR is quite important.

PR #2358, opened 2021-12-13
Release/3.36.2

tx0c:
release "version": "3.36.2"

merge after thematters/matters-server#2386


tx0c:
accidentally created against `develop`, should be #2359 merge toward `master`

PR #2364, opened 2021-12-15
feat(homepage): add feed type to query string instead of storing on local state

robertu7:


devformatters (approved):
It's a good start by using query string.

PR #2394, opened 2022-01-21
fix(copy): add and revise wallet integration copies

devformatters:


robertu7 (approved):
Guess we can use <Form.Field.Footer> outside of <Form.List.Item>

devformatters:
> Guess we can use <Form.Field.Footer> outside of <Form.List.Item>

I've tried it. But `Form.Field.Footer`'s padding & margin works differently.

PR #2408, opened 2022-01-26
remove the extra ConnectWalletButton

tx0c:


src/views/Me/Settings/ConnectWallet/index.tsx
```diff
@@ -19,6 +19,7 @@ const ConnectWallet = () => {
           submitCallback={() => {
             forward('wallet-connect')
           }}
+          back={() => history.back()}
```
robertu7, line 22: `router.back()`?

src/views/Home/index.tsx
```diff
@@ -46,13 +45,11 @@ const Home = () => {
                 <Layout.Header.Title id="discover" />
               )}
 
-              <section className="buttons">
```
robertu7, line 49: Based on design, we should hide the `<ConnectWalletButton>` in `<SortBy>` on mobile?

robertu7 (changes requested):

src/views/Me/Settings/ConnectWallet/index.tsx
tx0c: just found the `BackButton` already has such `router.back` logic but does not work, fix there instead

guoliu (approved):

PR #2411, opened 2022-01-27
chore(deps): update node version to 12.22

robertu7:
> The minimum Node.js version has been bumped from 12.0.0 to 12.22.0, which is the first version of Node.js with native ES modules support.
>
> https://nextjs.org/blog/next-12


devformatters (approved):
Upgrade. 💍

PR #2413, opened 2022-01-29
fix: show original error in error toasts if no translation found

Carolusian:


src/components/GQL/error.tsx
```diff
@@ -56,11 +56,11 @@ export const mutationOnError = (
 
   // Get error code and check corresponding content, if it's invalid
   // then expose error code
-  const errorCode = errorCodes[0] || ''
-  const errorContent = isErrorCodeValid(errorCode) ? (
-    <Translate id={errorCode} />
+  const errorCodeOrMessage = errorCodes[0] || error.message
```
robertu7, line 59: check if `TEXT.zh_hant[errorCode[0]]` exists first then fallback to `error.message`?

guoliu: Agree that if there is no translation to the code, showing the error code is better. But if there is `TEXT.zh_hant[errorCode[0]]`, it should be displayed to the user.

Carolusian: @robertu7 @guoliu updated accordingly, pls check.

robertu7 (approved):

PR #2414, opened 2022-02-12
fix/duplicate-tos: show the translations only when lang is not 'en'

### LINKED ISSUE #2376: TOS shows duplicate of terms and privacy policy, under English environment
png

image

tx0c:
the 'en' terms & privacy are always shown.

resolves #2376
TOS shows duplicate of terms and privacy policy, under English environment


guoliu (approved):

devformatters (approved):
Go. 💍

PR #2428, opened 2022-03-07
feat: sync with `generateSigningMessage` breaking change

robertu7:


tx0c (commented):
is there a connected PR with changes to `matters-server`?

and is there a standard on GraphQL API design? what's the benefit if there's only 1 parameter and need to be wrapped inside `input` ?

robertu7:
> is there a connected PR with changes to `matters-server`?

`generateSigningMessage` was changed on https://github.com/thematters/matters-server/pull/2458

> and is there a standard on GraphQL API design? what's the benefit if there's only 1 parameter and need to be wrapped inside `input` ?

all params wrapped inside `input` will backward compatible when new param is added. In this case, adding `input.purpose` won't break the previous usage if we used `input` at the beginning.

tx0c (approved):
I see;  so write this thematters/matters-server#2458 into commit message would make it clear since beginning

Issue #2436, opened 2022-03-23
[Web] - Update tag selector


devformatters:
tasks
- ~~Add default tag recommendation into tag selector~~
- Show each tag's articles total number in tag selector
- Disallow `#` and ` ` when creating new tag

design
- [figma](https://www.figma.com/file/oGew0i2s0N8IBm9tCpoLLI/%E8%BF%AD%E4%BB%A3%E8%A8%AD%E8%A8%88%E4%B8%AD%E7%B9%BC%E7%AB%99?node-id=3631%3A18098)

PR #2455, opened 2022-04-14
fix(): move strip logic from SearchInput back to SearchingArea/ for b…

tx0c:
…etter debounce control

for #2436


devformatters (approved):
Go. 🦾

PR #2479, opened 2022-04-22
Origin/fix/update english copy

ElkLeung:


tx0c (approved):
let's merge it before growing too big; and you can always make a part ii PR if not complete

Issue #2498, opened 2022-05-22
[Web] - Display issue of the manage button in tag detail page


lilac:
**Describe the bug**
When  a maintainer goes to a tag detail page, there is a "manage" button on the "community" tab. This button displays correctly if the language is Chinese, but if it's English, it's too narrow to display the whole word.

**To Reproduce**
1. Change the language to English.
2. Go to a tag detail page as the maintainer of the tag.
3. Click on the community tab.
4. The "manage" button display badly.

**Expected behavior**
The button should have appropriate size.

**Screenshots**
<img width="703" alt="Screen Shot 2022-05-22 at 9 20 17 PM" src="https://user-images.githubusercontent.com/705652/169697659-a13973f1-8313-400b-b0f9-eb5b422b73d8.png">

**Additional context**
It occurs on both desktop and phones.

devformatters:
@lilac 
Thanks for reporting this issue. I've picked it into our fixing process.

PR #2499, opened 2022-05-22
Make the size of manage button adaptive

lilac:
Fix this bug https://github.com/thematters/matters-web/issues/2498.

With this change, now the button look consistent.
image


guoliu (approved):
Thank you for submitting PR @lilac ! It looks good to me ❤️🌱

Issue #2518, opened 2022-06-07
[Web] Adjust button size


devformatters:
refs
- [slack](https://mattersnews.slack.com/archives/GA0R09FHN/p1653996683502079)

PR #2545, opened 2022-06-24
fix/adjust-follow-button-size: change the size of the followed/unfollo…

Jasmine-liang:
The spacing of unfollow/followed button is small due to the length of word, [issue link](https://github.com/thematters/matters-web/issues/2518)


robertu7 (approved):
🚀

PR #2560, opened 2022-06-29
use interval timer polling iscn publishing status

tx0c:
since iscn publishing took longer than expected (10-20s), for a total up to 30s;


robertu7 (approved):

tx0c:
cherry-pick'ed in fe10c37 instead

Issue #2348, opened 2021-12-08
[Web] Better support for tag URL and SEO


tx0c:
tasks
- [x] add `/tags/{id}-{slugified(tag-content)}` as entry of each tag
- [x] add SEO headers in tag page

PR #2561, opened 2022-06-29
feat(tags): use id-slug for tag path, and append `#feed` for tab afte…

tx0c:
…r initial

### DIFF
--- a/src/views/TagDetail/index.tsx
+++ b/src/views/TagDetail/index.tsx
@@ -1,5 +1,5 @@
-import _find from 'lodash/find'
-import _some from 'lodash/some'
+// import _find from 'lodash/find'
+// import _some from 'lodash/some'
 import dynamic from 'next/dynamic'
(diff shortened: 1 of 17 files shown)


src/views/TagDetail/index.tsx
```diff
@@ -60,16 +66,54 @@ const TagDetail = ({ tag }: { tag: TagDetailPublic_node_Tag }) => {
...
+    router.events.on('hashChangeStart', onHashChangeStart)
+
+    return () => {
+      router.events.off('hashChangeStart', onHashChangeStart)
+    }
+  }, [router.events])
+
+  useEffect(() => {
+    if (!router.isReady) return
+    router.asPath
```
robertu7, line 92: seems L75-L93 are unnecessary?

robertu7 (changes requested):

tx0c:
will close #2348 TBD in commit msg

PR #2568, opened 2022-07-02
HotRelease/4.3.5: edit iscn

tx0c:
fix(polling-iscn): use usePublicQuery with skip first instead of lazy……Query

- use interval timer polling iscn publishing status
- since iscn publishing took longer than expected (10-20s), for a total up to 30s;


robertu7 (approved):
LGTM

PR #2588, opened 2022-07-11
fix: fix styles of tag card

ElkLeung:


tx0c:
I have cherry-pick and tested on https://web-next.matters.news/ looking good

PR #2602, opened 2022-07-19
feat(downvote): remove downvote counts and query field

Jasmine-liang:
Someone have already removed the downvote counts and merged it into `develop` branch.
So I'm wondering how to remove **only** the downvote counts query field, been trying...


tx0c:
good try in your first commit 604de4f1ef34, but there's more use of `downvote` than changes in that commit,

you can try remove all occurrences of `downvote` ?

Issue #2590, opened 2022-07-13
[Web] SSR Issue


devformatters:
### Tasks
- [slack](https://mattersnews.slack.com/archives/GA0R09FHN/p1657425201760809)

PR #2606, opened 2022-07-20
fix: server-side rendering issue after next@12.1.5

Carolusian:
fix https://github.com/thematters/matters-web/issues/2590


Carolusian:
The SSR did not work starts from https://github.com/thematters/matters-web/commit/fd6cd7daa6a74bd85f4da1012e6e3dfd7deec236 , even downgrade to 12.1.5 and 12.1.6, the SSR still don't work. But 12.1.6-canary.3 works. See also: https://github.com/vercel/next.js/issues/36198

robertu7 (approved):
Let's give it a try!

tx0c:
it seems failed the EB deployment at https://web-develop.matters.news/

PR #2607, opened 2022-07-20
Release/4.3.7

tx0c:
3c5e258 fix(copy): update copies
75063b7 feat(tag): add <TagDigest.Sidebar>
530a734 fix(random): in range [0..50) but not including 50
86a446a fix(comment-url): hold fragment if there's any
4f21dfe feat(utm-tracking): better computing links source & destination before or after visiting a link


robertu7 (approved):
🚀

PR #2610, opened 2022-07-21
feat(tag): remove recommended tags from tag selector

robertu7:
Also change tag `<SearchSelectDialog>` to smaller width.


devformatters (approved):
Let's try it in dev env

PR #2621, opened 2022-07-26
feat(downvote): remove query field of downvote counts

Jasmine-liang:
Temporarily set the downvote counts to 0 and remove the query field from the response.

Met an error: `[GraphQL error]: Message: Fragment "DownvoteCommentPrivate"` is never used


src/components/Comment/FooterActions/index.tsx
```diff
@@ -143,9 +140,9 @@ const BaseFooterActions = ({
   return (
     <footer
       aria-label={translate({
-        zh_hant: `${comment.upvotes} 點讚、${comment.downvotes} 點踩`,
-        zh_hans: `${comment.upvotes} 点赞、${comment.downvotes} 点踩`,
-        en: `${comment.upvotes} upvotes, ${comment.downvotes} downvotes`,
+        zh_hant: `${comment.upvotes} 點讚、${0} 點踩`,
+        zh_hans: `${comment.upvotes} 点赞、${0} 点踩`,
+        en: `${comment.upvotes} upvotes, ${0} downvotes`,
```
tx0c, line 145: just remove all of `${comment.downvotes} downvotes` ... don't leave a `0` there

src/components/Comment/FooterActions/UpvoteButton/index.tsx
```diff
@@ -60,7 +59,7 @@ const UpvoteButton = ({
       unvoteComment: {
         id: comment.id,
         upvotes: comment.upvotes - 1,
-        downvotes: comment.downvotes,
+        downvotes: 0,
```
tx0c, line 62: just comment out, or remove the whole line, don't need a `0`

Jasmine-liang: Tired commenting out, the `downvotes` field is required. Discussed with Robert, could leave it as 0(could not be null also). Any thoughts?

robertu7 (approved):
LGTM

tx0c:
seems caused any circle broadcast unable to load

PR #2622, opened 2022-07-26
feat(track-exp): add delay_msecs to track delays of exposure events

tx0c:
for both CardExposure and TagExposure


guoliu (approved):
Nice!

PR #2631, opened 2022-07-28
fix: missing article id in sharing dialog

ElkLeung:


robertu7:
resolved by https://github.com/thematters/matters-server/pull/2651

Issue #2430, opened 2022-03-10
[Web] Return helpful message when wallet-registered user try to use email login


devformatters:
### Tasks
- [ ] Return helpful message only if email used in wallet-registered login
  - `當初是否透過錢包註冊？若是的話，請直接使用加密錢包登入，信箱僅作為聯繫渠道`
  - `当初是否透过钱包注册？若是的话，请直接使用加密钱包登入，信箱仅作为联系渠道`

### References
- [slack](https://mattersnews.slack.com/archives/GA0R09FHN/p1645629783906629?thread_ts=1643359894.738409&cid=GA0R09FHN)

### Screenshot
<img width="498" alt="截圖 2022-07-13 下午11 27 19" src="https://user-images.githubusercontent.com/34237112/178771950-5df6992d-8f2f-4d13-94df-7f970a8f9586.png">

PR #2679, opened 2022-08-07
fix: login error message for wallet registered users

Carolusian:
Fix https://github.com/thematters/matters-web/issues/2430


robertu7 (approved):
Do we also need some changes on server-side?

Carolusian:
> Do we also need some changes on server-side?

@robertu7 the server side shall be ready in this commit: https://github.com/thematters/matters-server/commit/58892b69207d16d2db23690cf88a01b1456c8f22

PR #2680, opened 2022-08-08
feat(url): use `?type=` for tag detail feeds

robertu7:
Also,
* removed some redundant code;
* added comments about backward compatibility;

related PR: https://github.com/thematters/matters-web/pull/2561


devformatters (approved):
Go. 🔧

PR #2690, opened 2022-08-15
feat: replace process.browser with typeof window

robertu7:
Base branch is `feat/deprecate-web-push`


robertu7:
Will open another PR

PR #2691, opened 2022-08-15
fix(relate-tags): send with trackRelatedTags instead

tx0c:
- fix with the function call
- and, avoid the confusing onClick meta function naming


robertu7 (approved):

Jasmine-liang:
Already added this chanegs into my new [add shuffle button into related tags](https://github.com/thematters/matters-web/pull/2693) PR :)

PR #2693, opened 2022-08-15
feat(related-tags): add shuffle button and function into related tags

Jasmine-liang:
### Changes
- Add Shuffle button component
- Add `random` input in `TagDetailRecommended` query

### DIFF
--- a/src/views/TagDetail/RelatedTags/index.tsx
+++ b/src/views/TagDetail/RelatedTags/index.tsx
@@ -1,10 +1,13 @@
+import { useQuery } from '@apollo/react-hooks'
 import classNames from 'classnames'
 import _chunk from 'lodash/chunk'
 import _get from 'lodash/get'
+import _random from 'lodash/random'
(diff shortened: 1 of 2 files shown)


src/views/TagDetail/RelatedTags/index.tsx
```diff
@@ -74,7 +107,9 @@ const RelatedTags: React.FC<RelatedTagsProps> = ({ tagId, inSidebar }) => {
   if (!inSidebar) {
     return (
       <section className={relatedTagsClasses}>
-        <Slides header={<RelatedTagsHeader />}>
+        <Slides
+          header={<RelatedTagsHeader hasShuffle={true} onShuffle={shuffle} />}
```
tx0c, line 111: you can omit the `{true}` for booleans

Jasmine-liang: Thanks!

PR #2695, opened 2022-08-15
fix(relate-tags): send with trackRelatedTags instead

tx0c:
- fix with the function call, from previous problem was a meta function not calling
- and, avoid the confusing onClick meta function naming


tx0c:
replaced by #2696

PR #2697, opened 2022-08-15
fix(relate-tags): send with trackRelatedTags instead

tx0c:
* dbc561967 (HEAD) feat(web-next): turn noindex to bots for web-next
* 4393e5eda (origin/release/4.3.8, local) Update Meson gateway with new API
* 9690210cc (origin/release/relate-tags-) fix(relate-tags): send with trackRelatedTags instead

- fix with the function call, from previous problem was a meta function not calling
- and, avoid the confusing onClick meta function naming

PR #2698, opened 2022-08-15
Update Meson gateway with new API

guoliu:
As title.


tx0c:
pick in #2697 in 4393e5e instead

PR #2703, opened 2022-08-18
fix: name length limit

ElkLeung:
add props in `<Tag>` & `<UserDigest.Rich>`


robertu7 (changes requested):
`<Tag>` and `<UserDigest.Rich>` are used in many places besides the sidebar, I think hardcoded isn't an ideal solution in this case. If must, shall we add a new optional component prop to only apply to sidebar?

BTW, which issue is this PR fix for? Could you link it to this PR?

robertu7 (approved):
LGTM.

PR #2711, opened 2022-08-22
fix(page-referrer): should be a full URL, not only pathname

tx0c:
otherwise, recognized as external links in Analytics Console


robertu7 (approved):
LGTM

robertu7:
@tx0c shall we merge it?

tx0c:
already in 137dde7e14 in #2710

PR #2720, opened 2022-08-22
fix: limit length rule

ElkLeung:


robertu7 (changes requested):
Use `tag.numberArticles ? <...> : null` might fix this issue.

Issue #2716, opened 2022-08-22
Show the same result when switching tag feeds


devformatters:
### Tasks
- [ ] Investigate and fix

devformatters:
Close it since PR was merged.

PR #2727, opened 2022-08-25
*[master]* hoxfix: show the same result when switching tag feeds

pitb2022:
https://github.com/thematters/matters-web/issues/2716


robertu7 (approved):

src/views/TagDetail/Articles/index.tsx
```diff
@@ -171,7 +171,7 @@ const TagDetailArticles = ({ tagId, selected }: TagArticlesProps) => {
       <List>
         {(edges || []).map(({ node, cursor }, i) => (
           <>
-            <List.Item key={cursor}>
+            <List.Item key={`${selected ? 'selected' : 'latest'}:${cursor}`}>
```
tx0c, line 174: looks good, but why not same as in #2728 

since just released yesterday; the master and develop should be very same or similar

PR #2728, opened 2022-08-25
fix:Show the same result when switching tag feeds

pitb2022:
https://github.com/thematters/matters-web/issues/2716


robertu7 (approved):
LGTM

PR #2746, opened 2022-09-02
feat(i18n): auto translate article content

robertu7:
### Changes
* auto translate article contents if URL has language subpath;
* insert language subpath to article sharing URL if article contents are translated;

### Notes
* Base branch is `feat/i18n-config`;


src/views/ArticleDetail/Toolbar/index.tsx
```diff
@@ -62,9 +64,22 @@ const fragments = {
...
+  translated,
+  translatedLanguage,
+  privateFetched,
+  lock,
+  ...props
+}: ToolbarProps) => {
   const isSmallUp = useResponsive('sm-up')
 
+  const path = toPath({ page: 'articleDetail', article })
+  const shareUrl =
```
pitb2022, line 78: shareUrl needs to be passed in on the mobile side

robertu7: resolved by https://github.com/thematters/matters-web/pull/2764

Issue #2694, opened 2022-08-15
Update about us page


devformatters:
### Tasks
- [ ] Update about us page content
- [ ] Update styles
- [ ] change footer `Matters 長什麼樣` to `關於我們`

### References
- [Figma](https://www.figma.com/file/NWPnD0xUHxz1fx0xzNCWLc/About-Us-%E9%97%9C%E6%96%BC%E6%88%91%E5%80%91?node-id=1%3A2245)
- [Copies](https://mattersnews.slack.com/archives/G805NQVE3/p1659331118422079)

PR #2748, opened 2022-09-05
Feat/update about us

pitb2022:
https://github.com/thematters/matters-web/issues/2694


src/views/About/Team/index.tsx
```diff
@@ -81,31 +85,64 @@ const CREW = [
...
-      },
       {
         avatar: IMAGE_ROBERT,
-        title: 'Product Engineer',
+        title: 'Tech Lead',
         name: 'Robert Lu',
       },
+      {
+        avatar: IMAGE_TOMAS,
+        title: 'Full Stack Engineer',
```
robertu7, line 95: this isn't synced with doc?

robertu7 (changes requested):

src/views/About/Team/index.tsx
pitb2022: 👌

PR #2771, opened 2022-09-12
fix(tag): allow re-use old existing non-standard tag names

tx0c:


robertu7 (approved):
Cannot find the changes about "allow re-use old existing non-standard tag names"...others LGTM.

PR #2773, opened 2022-09-13
Fix/web dev test

pitb2022:
As title


src/components/Dialogs/ShareDialog/Content.tsx
```diff
@@ -27,61 +28,63 @@ const ShareDialogContent: React.FC<ShareDialogContentProps> = ({
...
-    />
-
-    <Dialog.Content>
-      {description && (
-        <section className="description">
-          {description}
+}) => {
+  const url = new URL(shareLink)
+  const pathnames = url.pathname.split('/')
+  const showTranslation = toLocale(pathnames[1]) !== ''
```
robertu7, line 34: Could use this instead?

```tsx
import { DEFAULT_LOCALE } from '~/common/enums'

...

const router = useRouter()
const showTranslation = router.locale !== DEFAULT_LOCALE
```

robertu7 (approved):

src/components/Dialogs/ShareDialog/Content.tsx
pitb2022: Doesn't seem to work!

PR #2790, opened 2022-09-20
feat(analytics): data collection of try-login from visitor

tx0c:


src/views/ArticleDetail/Wall/Visitor/index.tsx
```diff
@@ -59,7 +57,15 @@ const VisitorWall = ({ show }: VisitorWallProps) => {
           </p>
 
           <div className="signup">
-            <LoginButton bgColor="green" />
+            <LoginButton
```
robertu7, line 60: Besides here, there are other components/pages using `<LoginButton>`, do they need this tracking?

robertu7 (approved):

PR #2791, opened 2022-09-21
feat(rss): add RSS feed UI

Jasmine-liang:
Add RSS feed icon to user profile; 
Add RSS dialog;


robertu7:
image

need a hover/focus effect here.

src/common/utils/resolvers/rssFeedGatewayUrls.ts
```diff
@@ -0,0 +1,94 @@
+import fetch from 'isomorphic-unfetch'
```
robertu7, line 1: i think we can reuse `gatewayUrls`.

tx0c:
rebase and squash ? into single commit as 26dc4dcc5 testing in `release/next`

src/components/Dialogs/RssFeedDialog/Content.tsx
```diff
@@ -0,0 +1,210 @@
...
+                </li>
+              )
+            })}
+          </ul>
+        </section>
+
+        <hr style={{ margin: '1rem 0' }} />
+
+        {/* hash */}
+        <section className="hash">
```
tx0c, line 159: the whole section should be conditionally displayed, aka. don't show it when there's no `ipnsKey`

```js
{ipnsKey && <section ...
}
```

PR #2793, opened 2022-09-23
[hot fix]fix(copies): correct english copies

Jasmine-liang:
As title


robertu7:
Same question as https://github.com/thematters/matters-web/pull/2794#issuecomment-1255945340

robertu7 (approved):

PR #2806, opened 2022-09-30
feat: revise Wallet

pitb2022:
As Title


robertu7 (changes requested):
Left some comments about the UI on Slack thread, please confirm with designer, thanks.

PR #2811, opened 2022-10-06
feat(rss):add content feed UI

Jasmine-liang:


devformatters (approved):
Let's try it on dev env

robertu7 (changes requested):

src/components/UserProfile/index.tsx
```diff
@@ -297,3 +321,5 @@ export const UserProfile = () => {
     </>
   )
 }
+
+export default RssFeedButton
```
devformatters, line 325: Why export Button as default ? 🤔

PR #2895, opened 2022-11-09
fix(use-id): use articleNodeId to query responses instead

tx0c:
fix(fallback-empty-string): fallback mediaHash/dataHash to '' empty s… related thematters/matters-server#2912


robertu7 (approved):
LGTM

PR #2910, opened 2022-11-11
fix(SupportWidget): trigger query HasDonated when the user is logged in

pitb2022:


src/views/ArticleDetail/SupportWidget/index.tsx
```diff
@@ -50,21 +50,30 @@ const SupportWidget = ({ article }: DonationProps) => {
...
+  useEffect(() => {
+    if (viewer.id !== '') {
+      getHasDonated({
+        variables: {
+          id: article.id,
+          senderId: viewer.id,
+        },
+      })
+    }
+  }, [viewer])
```
robertu7, line 69: use `viewer.id`

src/views/ArticleDetail/SupportWidget/index.tsx
```diff
@@ -50,21 +50,30 @@ const SupportWidget = ({ article }: DonationProps) => {
...
+  const [
+    getHasDonated,
+    { data: hasDonatedData, loading, refetch: hasDonatedRefetch },
+  ] = useLazyQuery<HasDonated>(HAS_DONATED, {
     fetchPolicy: 'network-only',
-    variables: { id: article.id, senderId: viewer.id },
   })
 
+  useEffect(() => {
+    if (viewer.id !== '') {
```
robertu7, line 61: return-early

```
if (!viewer.id) return 

...
```

src/views/ArticleDetail/SupportWidget/index.tsx
```diff
@@ -50,21 +50,30 @@ const SupportWidget = ({ article }: DonationProps) => {
...
+    }
+  }, [viewer])
+
   const hasDonatedArticle =
     hasDonatedData?.article as HasDonated_article_Article
 
   useEffect(() => {
     if (hasDonatedData) {
-      if (viewer.id !== '' && hasDonatedArticle?.donation.totalCount === 1) {
+      if (hasDonatedArticle?.donation.totalCount === 1) {
```
robertu7, line 76: `hasDonatedArticle?.donation.totalCount >= 1`?

robertu7 (changes requested):

src/views/ArticleDetail/SupportWidget/index.tsx
pitb2022: This request is only for the current user, so it must be equal to one

robertu7 (approved):

PR #2916, opened 2022-11-14
Feat/currency converter preference

pitb2022:
https://github.com/thematters/matters-web/issues/2867

### DIFF
--- a/src/views/Me/Settings/Settings/UI/CurrencyConvertor/index.tsx
+++ b/src/views/Me/Settings/Settings/UI/CurrencyConvertor/index.tsx
@@ -107,25 +107,15 @@ const CurrencyConvertor = () => {
       }}
       dialog={{
         content: <Content />,
-        // TODO: comfirm title with PM
         title: (
-          <Translate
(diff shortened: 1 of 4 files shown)


src/views/Me/Settings/Settings/UI/CurrencyConvertor/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  return (
+    <DropdownDialog
+      dropdown={{
+        content: <Content isInDropdown />,
+        placement: 'bottom-end',
+      }}
+      dialog={{
+        content: <Content />,
+        title: (
+          <Translate zh_hant="修改幣種" zh_hans="修改币种" en="Currency" />
```
robertu7, line 111: Not the same as L118?

robertu7 (approved):

src/views/Me/Settings/Settings/UI/CurrencyConvertor/index.tsx
pitb2022: done!

PR #2923, opened 2022-11-15
feat(ipns): make IPNS entrance public

robertu7:


src/components/UserProfile/index.tsx
```diff
@@ -255,9 +255,7 @@ export const UserProfile = () => {
           <section className="right">
             {!isMe && <FollowUserButton user={user} size="lg" />}
 
-            {totalCount !== 0 && hasTraveloggersBadge && (
-              <RssFeedButton user={user} />
-            )}
+            {totalCount > 0 && <RssFeedButton user={user} />}
```
tx0c, line 258: still need another condition to show it only for those we have ipns_key

    totalCount > 0 && user.info.ipnsKeyey &&

tx0c (approved):

PR #2963, opened 2022-12-01
fix(csp-access): revise nft avatar image url

Jasmine-liang:


src/components/UserProfile/DropdownActions/EditProfileDialog/NFTCollection/index.tsx
```diff
@@ -37,7 +37,10 @@ const NFTCollectionItem = ({
       onClick={onClick}
       title={`${name}\n- ${description}`}
     >
-      <Avatar size="xxxl" src={imagePreviewUrl} />
+      <Avatar
+        size="xxxl"
+        src={`https://server-develop.matters.news/img-cache/${imagePreviewUrl}`}
```
robertu7, line 42: should not hardcode

robertu7: and production is not this domain

PR #2967, opened 2022-12-05
feat(ens-poc): bind IPNS hash to ENS name

Jasmine-liang:
Make sure your network is on Goerli and then can test the set contenthash function.
To be concerned:
1. Here we use `PublicResolver` abi. Not sure each ENS shares this default abi on mainnet.


src/components/UserProfile/index.tsx
```diff
@@ -87,6 +112,27 @@ export const UserProfile = () => {
     })
   }, [user?.id, viewer.id])
 
+  const bindEnsIpns = async () => {
+    if (!eip1577 || !ensName) {
+      return
+    }
+    const ipnsHash = user?.info.ipnsKey
+    const ensNameHash = namehash(ensName as string)
+    const encoded = '0x' + contentHash.encode('ipns-ns', ipnsHash)
```
robertu7, line 121: I found the package size of `@ensdomains/content-hash` is [pretty big](https://bundlephobia.com/package/@ensdomains/content-hash@2.5.7), could we remove the dep if we only use `encode`?

src/components/UserProfile/index.tsx
```diff
@@ -63,6 +74,21 @@ export const UserProfile = () => {
   const { getQuery } = useRoute()
   const viewer = useContext(ViewerContext)
 
+  const { address } = useAccount()
+  const { data: ensName } = useEnsName({
+    address,
+  })
+  const { data: signer } = useSigner()
+  const { data: resolverData } = useEnsResolver({
```
robertu7, line 82: So `useEnsResolver ` can query the resolver address of provided ENS name? 🐸

src/common/utils/ensResolver.ts
```diff
@@ -0,0 +1,214 @@
...
+    stateMutability: 'nonpayable',
+    type: 'function',
+  },
+  {
+    constant: false,
+    inputs: [
+      { name: 'node', type: 'bytes32' },
+      { name: 'hash', type: 'bytes' },
+    ],
+    name: 'setContenthash',
```
robertu7, line 58: Looks like we only use this method, could we remove other lines of this ABI file?

src/components/UserProfile/index.tsx
```diff
@@ -63,6 +74,21 @@ export const UserProfile = () => {
...
+    name: ensName as string,
+  })
+  // ENS related hooks
+  const eip1577 = useContract({
```
robertu7, line 86: Shall we have to confirm if all resolvers follow EIP1577 spec and have the `setContenthash`?

robertu7 (changes requested):

src/components/UserProfile/index.tsx
Jasmine-liang: Yes, tested on Goerli and Mainnet.

PR #2972, opened 2022-12-08
Feat/search history

pitb2022:


src/views/Search/index.tsx
```diff
@@ -21,11 +26,34 @@ import SearchTags from './SearchTags'
 import SearchUsers from './SearchUsers'
 
 const Search = () => {
+  const viewer = useContext(ViewerContext)
+  const storageKey = STORAGE_KEY_SEARCH_HISTORY + '_' + viewer.id
```
robertu7, line 30: please confirm if we should store the search history from anonymous

robertu7 (approved):

src/views/Search/index.tsx
pitb2022: Product doc: [login users should have independent history, anonymous doesn’t matter](https://www.notion.so/matterslab/Search-Rework-c66bd1db9c9c49d8a5a4b22cff5244b7#2977386c9bd04dd39735707ba9c442f0)

PR #2987, opened 2022-12-16
feat(Search): add quick result hotkeys

pitb2022:


src/components/Search/SearchQuickResult/index.tsx
```diff
@@ -34,8 +37,15 @@ export const SearchQuickResult = (props: QuickSearchProps) => {
   const hasUsers = userEdges && userEdges.length > 0
   const hasTags = tagEdges && tagEdges.length > 0
 
+  useEffect(() => {
+    if (onUpdateData) {
+      onUpdateData(data)
+    }
+  }, [data])
```
robertu7, line 44: Need to confirm if `useEffect` always got fired since `[data]` isn't a deep comparison? 

ref: 
https://medium.com/suyeonme/react-lets-deep-dive-into-deps-array-of-useeffect-13ab96468db7
https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

robertu7 (approved):

PR #3020, opened 2022-12-28
fix(ens): add wagmi global cache config

Jasmine-liang:


robertu7 (approved):
can add a `FIXME` comment for further fix

PR #3032, opened 2022-12-30
fix(supporter-list): replace InfiniteList to InfiniteScroll

Jasmine-liang:


src/components/Dialogs/DonatorsDialog/Content.tsx
```diff
@@ -22,6 +22,7 @@ import {
   ArticleDonators_article_Article_donations_edges,
 } from './__generated__/ArticleDonators'
 import { DonatorDialogArticle } from './__generated__/DonatorDialogArticle'
+// import { List } from 'react-virtualized'
```
robertu7, line 25: can remove `react-virtualized` and `@types/react-virtualized`?

src/components/Dialogs/DonatorsDialog/Content.tsx
```diff
@@ -129,17 +115,24 @@ const DonatorsDialogContent = ({
...
-        </div>
+        <InfiniteScroll
+          loader={<Spinner />}
+          loadMore={loadMore}
+          hasNextPage={pageInfo.hasNextPage}
+        >
+          {edges.map(({ node, cursor }, i) => (
+            <List.Item key={cursor}>
+              <div className="dialog-donator-list">
+                <ListRow
```
robertu7, line 126: shall we just move `<UserDigest.Rich>` here instead of using `<ListRow>`? Also, the `.dialog-donator-list` and `.donator-item` seems unused.

src/components/Dialogs/DonatorsDialog/Content.tsx
Jasmine-liang: `AppreciatorsDialogContent` [component](https://github.com/thematters/matters-web/blob/49e3a2cb1ade40d45b8950df788b05f7d6f9b891/src/components/Dialogs/AppreciatorsDialog/Content.tsx#L6) still uses `InfiniteList`, which uses `react-virtualized` pacakge🤔. Shall we refactor `AppreciatorsDialogContent` as well?

src/components/Dialogs/DonatorsDialog/Content.tsx
Jasmine-liang: Now `.dialog-donator-list` is adding padding

robertu7 (approved):

PR #3033, opened 2022-12-30
fix(support-animation): direct to SupportWidget animation after donation

Jasmine-liang:


src/views/ArticleDetail/SupportWidget/index.tsx
```diff
@@ -79,6 +81,12 @@ const SupportWidget = ({ article }: DonationProps) => {
     }
   }, [isViewerDonated])
 
+  useEffect(() => {
+    if (showAnimation && animationRef?.current) {
+      animationRef?.current.scrollIntoView({ behavior: 'smooth' })
```
robertu7, line 86: use `jump()`?

src/views/ArticleDetail/SupportWidget/index.tsx
```diff
@@ -79,6 +81,12 @@ const SupportWidget = ({ article }: DonationProps) => {
     }
   }, [isViewerDonated])
 
+  useEffect(() => {
```
robertu7, line 84: why not just add it to [`useEventListener`](https://github.com/thematters/matters-web/blob/0a5f7d8906399d5cd9e8e1ab4e9bba7432fb5b2a/src/views/ArticleDetail/SupportWidget/index.tsx#L100), any difference?

robertu7 (approved):

PR #3043, opened 2023-01-04
feat(adsense-script): adsense NextScript

tx0c:


src/components/Head/index.tsx
```diff
@@ -218,6 +219,13 @@ export const Head: React.FC<HeadProps> = (props) => {
           key="ld-json-data"
         />
       )}
+
+      <NextScript
```
robertu7, line 223: https://nextjs.org/docs/basic-features/script should use `next/script`?

robertu7 (changes requested):

robertu7 (approved):

PR #3053, opened 2023-01-04
fix(support-animation): put jump() into use EventListener

Jasmine-liang:


robertu7 (changes requested):
please remove `__generated__` and rebase the branch (still using the old codegen script?)

PR #3097, opened 2023-01-18
WIP:feat(i18n): add format.js library

Jasmine-liang:


robertu7:
https://github.com/vercel/next.js/tree/canary/examples/with-react-intl can have a look if it's helpful.

compiled-lang/zh_hant.json
```diff
@@ -0,0 +1,14 @@
+{
```
robertu7, line 1: `compiled-lang` should be gitignored?

PR #3120, opened 2023-02-06
[hotfix - develop](Expandable): Compatible with Safari below 15

pitb2022:
Use clip instead of hidden in order to be compatible with Safari 15 and below

### DIFF
--- a/src/components/Expandable/styles.css
+++ b/src/components/Expandable/styles.css
@@ -14,7 +14,7 @@
   }
 
   & .richWrapper {
-    overflow-y: clip;
+    overflow-y: hidden;
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/components/Expandable/styles.css
```diff
@@ -14,7 +14,7 @@
   }
 
   & .richWrapper {
-    overflow-y: clip;
```
robertu7, line 17: Could we keep both of them?

```css
overflow-y: hidden;
overflow-y: clip;
```

robertu7 (changes requested):

src/components/Expandable/styles.css
pitb2022: 👍

robertu7 (approved):

PR #3124, opened 2023-02-07
feat(i18n): add translation example

Jasmine-liang:
Add translation examples covering all scenarios: text, text+variable, richtext


package.json
```diff
@@ -28,14 +28,18 @@
...
+    "i18n:compile": "formatjs compile-folder --ast lang compiled-lang",
+    "i18n": "npm run i18n:extract && npm run i18n:compile"
   },
   "dependencies": {
     "@apollo/react-common": "^3.1.3",
     "@apollo/react-hooks": "^3.1.3",
     "@apollo/react-ssr": "^3.1.3",
     "@artsy/fresnel": "^6.1.0",
     "@ensdomains/content-hash": "^2.5.7",
+    "@formatjs/cli": "^6.0.0",
```
robertu7, line 42: should be in `devDependencies`?

tsconfig.json
```diff
@@ -4,9 +4,13 @@
     "allowSyntheticDefaultImports": true,
     "jsx": "preserve",
     "lib": [
+      "esnext.intl",
```
robertu7, line 7: any references on these changes?

src/views/Guide/index.tsx
```diff
@@ -22,10 +24,16 @@ const Guide = () => {
...
         <section
           dangerouslySetInnerHTML={{
-            __html: translate({
-              ...content,
-              lang,
-            }),
+            // richtext translation
+            __html: intl.formatMessage(
+              {
+                defaultMessage: '{content}',
```
robertu7, line 30: IMO, `{content}` is confusing for users in Crowdin. As we discussed this afternoon, we may move this kind of long text to a single `lang` file. For now, we could skip this use case (keep `translate`).

robertu7 (changes requested):

tsconfig.json
Jasmine-liang: Typescript usage in `react-intl`: https://formatjs.io/docs/react-intl/

PR #3222, opened 2023-02-27
Fix/e2e test resouce loading order

pitb2022:


tests/authentication.spec.ts
```diff
@@ -2,11 +2,11 @@ import { expect, test } from '@playwright/test'
 
 import { TEST_ID } from '~/common/enums'
 
-import { authedTest, login, logout } from './helpers'
+import { authedTest, login, logout, pageGoto } from './helpers'
 
 test.describe('Authentication', () => {
   test('can login in homepage dialog', async ({ page, isMobile }) => {
-    await page.goto('/')
+    await page.goto('/', { waitUntil: 'networkidle' })
```
robertu7, line 9: use `pageGoto` as well?

robertu7 (approved):

tests/authentication.spec.ts
pitb2022: 👍

PR #3231, opened 2023-03-02
fix(TextIcon): fix incomplete font display error

pitb2022:


robertu7 (approved):
any context on this PR?

PR #3258, opened 2023-03-10
feat(Search): add analytics

pitb2022:


robertu7 (changes requested):
Please revise as we [discussed on Slack](https://mattersnews.slack.com/archives/C88CK7Q7L/p1678447257810199?thread_ts=1678433434.404789&cid=C88CK7Q7L)

robertu7 (changes requested):
pass search key on trackEvent?

PR #3319, opened 2023-03-31
Website UI Revision Pt.1

robertu7:
* revise homepage `<Announcements>`;
* border radius `<Cover>`;
* adjust subtext (digest footer, sidebar, etc.) font styles;
* fix cropped descender of letters;
* make unselected `<Tabs>` darken;


src/components/CircleDigest/Plain/styles.css
```diff
@@ -6,4 +6,8 @@ section {
 
 .name {
   @mixin line-clamp;
+
+  /* FIXME: fix cropped descenders of letters  */
```
robertu7, line 10: https://stackoverflow.com/questions/63977541/descenders-of-letters-get-cropped-out-when-using-webkit-line-clamp-with-reduced

pitb2022: Fix in <TextIcon />
https://github.com/thematters/matters-web/blob/develop/src/components/TextIcon/styles.css#L29#L39

pitb2022 (changes requested):

pitb2022 (approved):

PR #3323, opened 2023-04-03
Hotfix/revise few UI change

pitb2022:


src/components/Interaction/Card/styles.css
```diff
@@ -43,31 +41,6 @@
...
-
-  &:not(.disabled) {
-    &:hover,
-    &:focus {
-      background: var(--color-grey-lighter-active);
-    }
-  }
-}
-
-.bg-active-grey-lighter {
```
robertu7, line 55: This is a breaking change, right? Try global search `bgActiveColor="grey-lighter"`, there are many dependences.

PR #3332, opened 2023-04-11
Feat/revise image component

pitb2022:


.env.dev
```diff
@@ -7,7 +7,8 @@
 ###
 NEXT_PUBLIC_RUNTIME_ENV=development
 NEXT_PUBLIC_SITE_DOMAIN=web-develop.matters.news
-NEXT_PUBLIC_ASSET_DOMAIN=assets-develop.matters.news
+NEXT_PUBLIC_LEGACY_ASSET_DOMAIN=assets-develop.matters.news
```
robertu7, line 10: can also update to `.env.local.example`

src/common/utils/text/article.ts
```diff
@@ -100,9 +100,8 @@ export const optimizeEmbed = (content: string) => {
         return /* html */ `
       <picture>
         <source
-          type="image/webp"
```
robertu7, line 103: L102-L106 are duplicated with L108-L112

src/components/ResponsiveImage/index.tsx
```diff
@@ -35,12 +35,10 @@ const BaseResponsiveImage = ({
       {smUpSize && (
         <>
           <source
-            type="image/webp"
```
robertu7, line 38: Same duplication issue as above.

robertu7 (changes requested):

.env.dev
pitb2022: done

src/common/utils/text/article.ts
pitb2022: Removed

src/components/ResponsiveImage/index.tsx
pitb2022: Removed

PR #3355, opened 2023-04-14
feat(signup): turn off recaptcha check in signup page

gary02:


robertu7 (changes requested):
also need to revise/disable components that used `ReCaptchaContext`. e.g. https://github.com/thematters/matters-web/blob/develop/src/components/Forms/EmailSignUpForm/Init.tsx#L54

robertu7 (approved):

PR #3412, opened 2023-05-04
Feat/revise UI

pitb2022:


src/components/Avatar/styles.css
```diff
@@ -18,7 +18,7 @@
  * Ring
  */
 .hasRing {
-  margin: 2px 0;
+  margin: 2px 0 0 2px;
```
robertu7, line 21: left comment in Slack, please take a look

robertu7 (approved):

PR #3426, opened 2023-05-06
fix(SearchSelect): fix unable to search user on tag editor dialog

pitb2022:


robertu7 (commented):
image

Found a UI issue (overlap border) which not related to this bug.

PR #3432, opened 2023-05-08
fix(domain-text-links): change remaining .news text to .town

tx0c:
- 7bbbfa46a fix(domain-text-links): change remaining .news text to .town

### DIFF
--- a/src/views/TagDetail/index.tsx
+++ b/src/views/TagDetail/index.tsx
@@ -108,8 +108,8 @@ const TagDetail = ({ tag }: { tag: TagFragmentFragment }) => {
   // define permission
   const isOwner = tag?.owner?.id === viewer.id
   const isEditor = (tag?.editors || []).some((t) => t.id === viewer.id)
-  const isMatty = viewer.info.email === 'hi@matters.news'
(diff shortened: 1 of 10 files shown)


src/views/TagDetail/index.tsx
```diff
@@ -108,8 +108,8 @@ const TagDetail = ({ tag }: { tag: TagFragmentFragment }) => {
   // define permission
   const isOwner = tag?.owner?.id === viewer.id
   const isEditor = (tag?.editors || []).some((t) => t.id === viewer.id)
-  const isMatty = viewer.info.email === 'hi@matters.news'
-  const isMaintainer = isOwner || isEditor || isMatty
+  const isAdmin = viewer.status?.role === 'admin' // viewer.info.email === 'hi@matters.town'
```
robertu7, line 111: This change won't work for non-Matty admins since it's hardcoded Matty Id in backend.

robertu7 (approved):

src/views/TagDetail/index.tsx
tx0c: then should we better fix in the server side to allow all other 17 admins to be able edit tag?

PR #3434, opened 2023-05-08
fix(domain-text-links): change remaining .news text to .town

tx0c:


src/components/Forms/CreateCircleForm/styles.css
```diff
@@ -7,7 +7,6 @@
     left: var(--spacing-base);
     line-height: var(--input-height);
     color: var(--color-black);
-    content: 'matters.news/~';
```
robertu7, line 10: Shouldn't delete this?

PR #3488, opened 2023-05-26
[Hotfix - master] fix(Search): update search api

pitb2022:


robertu7:
can close since we will have a hot release later today

PR #3518, opened 2023-06-05
feat: add cc4.0 license

gary02:


src/components/Editor/ToggleAccess/SelectLicense/About/index.tsx
```diff
@@ -0,0 +1,17 @@
+import { IconExternalLink16 } from '~/components'
+
+import styles from './styles.css'
+
+const About = ({ desc, url }: { desc: string; url: string }) => {
+  return (
+    <>
+      <a href={url} target="_blank">
+        {desc}
```
robertu7, line 9: Can use `<TextIcon>` so you can save some CSS code.

robertu7 (changes requested):

robertu7 (approved):

PR #3565, opened 2023-06-14
Feat/article feed digest

pitb2022:


public/static/icons/18px/donate.svg
```diff
@@ -0,0 +1,3 @@
...
+</svg>
```
robertu7, line 3: can change the `fill` to `currentColor` so it can inherit the color from parent

robertu7 (changes requested):
Can add components to Storybook

robertu7 (approved):

PR #3609, opened 2023-07-14
Feat/Collection Detail Page

pitb2022:


src/views/User/CollectionDetail/CollectionArticles/index.tsx
```diff
@@ -0,0 +1,248 @@
...
+  let articleEdges = articles.edges
+  const [items, setItems] = useState(articleEdges)
+
+  useEffect(() => {
+    if (hasReset) {
+      setItems(articleEdges)
+    }
+  }, [articleEdges])
+
+  if (isViewer) {
```
robertu7, line 87: Can make it as standalone component and use dynamic import, so we can enhance the performance which only viewer (author) will load the  `react-movable` lib;

robertu7 (approved):

PR #3665, opened 2023-07-28
Fix/user page stage 9

pitb2022:


src/components/ArticleDigest/Feed/styles.module.css
```diff
@@ -70,4 +70,10 @@
   align-items: center;
   margin-bottom: var(--spacing-x-tight);
   line-height: 1;
+
+  & .author {
+    & :global(span[data-test-id='digest/user/mini/display-name']) {
+      line-height: 1.125rem;
```
robertu7, line 76: `data-test-id` is better only used for testing purposes.

robertu7: can we just modify the style of `<UserDigest.Mini>`?

robertu7 (approved):

robertu7 (approved):

PR #3689, opened 2023-08-01
Fix/user page stage 10

pitb2022:


src/components/Tabs/index.tsx
```diff
@@ -4,30 +4,48 @@ import Link from 'next/link'
...
-          {count && <span className={styles.count}>&nbsp;{count}</span>}
-        </a>
-      </Link>
+    <li
+      role="tab"
+      aria-selected={selected}
+      className={classes}
+      onClick={onClick}
+    >
+      <button>
```
robertu7, line 45: move `onClick` here for a11y

PR #3740, opened 2023-08-15
Feat/sign up with email

pitb2022:


src/components/Forms/EmailSignUpForm/Init.tsx
```diff
@@ -138,100 +107,96 @@ const Init: React.FC<FormProps> = ({
...
+        <FormattedMessage
+          defaultMessage="Continue"
+          description="src/components/Forms/EmailSignUpForm/Init.tsx"
+        />
+      }
       loading={isSubmitting}
     />
   )
 
-  if (isInPage) {
```
robertu7, line 181: Please confirm with product, do we still need to support login / sign up via page instead of dialog?

src/components/Forms/SelectAuthMethodForm/index.tsx
```diff
@@ -27,39 +28,29 @@ export const SelectAuthMethodForm: React.FC<FormProps> = ({
...
+        <AuthNormalFeed
+          gotoEmailSignup={gotoEmailSignup}
+          gotoEmailLogin={gotoEmailLogin}
+        />
+      )}
+      {isWallet && <AuthWalletFeed />}
     </>
   )
 
-  if (isInPage) {
```
robertu7, line 49: Same though as above about the compatibility with auth page.

robertu7 (changes requested):

robertu7 (approved):

PR #3771, opened 2023-08-28
Feat/sign in with social media

pitb2022:


src/common/enums/route.ts
```diff
@@ -120,6 +123,9 @@ export const ROUTES: {
   { key: 'LOGIN', pathname: '/login' },
   { key: 'SIGNUP', pathname: '/signup' },
   { key: 'FORGET', pathname: '/forget' },
+  { key: 'GOOGLE_CALLBACK', pathname: '/google-callback' },
```
robertu7, line 126: can only add one route instead: `/social-callback/[provider]`

robertu7 (changes requested):

robertu7 (approved):

PR #3794, opened 2023-09-07
feat(signup-referral): Emit GA event after account created

### LINKED ISSUE #3739: Emit GA event after account created

tx0c:
resolve #3739


src/views/Auth/Universal/index.tsx
```diff
@@ -76,9 +78,13 @@ const UniversalAuth = () => {
...
   useEffect(() => {
-    if (!viewer.id) return
-
-    router.push(PATHS.HOME)
+    if (viewer.id) {
+      router.push(PATHS.HOME)
+      return
+    }
+    if (referralCode) {
+      storage.set(STORAGE_KEY_REFERRAL_CODE, { referralCode })
```
robertu7, line 86: have to record the referralCode when user lands on `/login` and `/signup`

src/components/Buttons/UniversalAuth/index.tsx
```diff
@@ -23,6 +25,7 @@ export const UniversalAuthButton: React.FC = () => {
     onClick: () => {
       analytics.trackEvent('click_button', {
         type: 'login/signup',
+        referralCode: storage.get(STORAGE_KEY_REFERRAL_CODE)?.referralCode || undefined,
```
robertu7, line 28: emit the event once the [register is successful](https://github.com/thematters/matters-web/blob/master/src/components/Forms/EmailSignUpForm/Password.tsx#L127) is more accurate

robertu7 (changes requested):

src/views/Auth/Universal/index.tsx
tx0c: this `UniversalAuth` is used by both `/login` and `/signup`

robertu7 (approved):

PR #3805, opened 2023-09-12
Feat/revise auth page

pitb2022:


src/views/Auth/Universal/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+} from '~/components'
+
+import styles from './styles.module.css'
+
+const DynamicSelectAuthMethodForm = dynamic<any>(
+  () =>
+    import('~/components/Forms/SelectAuthMethodForm').then(
+      (mod) => mod.SelectAuthMethodForm
+    ),
+  { ssr: false, loading: Spinner }
```
robertu7, line 26: can turn on `ssr` so visitors won't notice the loading state.

robertu7 (approved):

PR #3824, opened 2023-09-19
Fix/new login stage 1

pitb2022:


src/views/Auth/Universal/index.tsx
```diff
@@ -79,82 +80,89 @@ const UniversalAuth = () => {
   }, [])
 
   return (
-    <section className={styles.wrapper}>
-      <section className={containerClasses}>
-        {currStep === 'select-login-method' && (
-          <>
-            <DynamicSelectAuthMethodForm
+    <>
+      <Head />
```
robertu7, line 84: better to have a title about authentication, please confirm with product team.

src/components/Dialogs/SetUserNameDialog/hook.ts
```diff
@@ -1,4 +1,5 @@
 import { useApolloClient } from '@apollo/react-hooks'
+import { toLower } from 'lodash'
```
robertu7, line 2: Use `lodash/toLower` for code splitting or just use `.toLowerCase()`

robertu7 (changes requested):
please double-check the dialog spacing issues

src/views/Auth/Universal/index.tsx
pitb2022: Confirm the use of the default title

robertu7 (approved):

PR #3873, opened 2023-10-06
fix(draft): create new draft if user tries to upload cover

robertu7:
related:
* https://github.com/thematters/matters-web/pull/3848
* https://github.com/thematters/matters-web/pull/3793


devformatters (approved):
Let's test it in dev env. 🌊

PR #3893, opened 2023-10-13
fix(useColorThief): fix theme color

pitb2022:
fix theme color of pinnedWork not updated after revision of article cover in Firefox 
ref:
https://feedback.matters.town/bug-reports/p/cover

### DIFF
--- a/src/components/Hook/useColorThief.tsx
+++ b/src/components/Hook/useColorThief.tsx
@@ -33,8 +33,13 @@ export const useColorThief = () => {
             2
           )}% 30%)`
         )
-      } catch (error) {
-        //
+      } catch (error: any) {
+        if (error.name === 'SecurityError') {
+          // Throws this error in Firefox
+          setTimeout(() => {
+            _getColor()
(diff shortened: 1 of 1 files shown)


src/components/Hook/useColorThief.tsx
```diff
@@ -33,8 +33,13 @@ export const useColorThief = () => {
             2
           )}% 30%)`
         )
-      } catch (error) {
-        //
+      } catch (error: any) {
+        if (error.name === 'SecurityError') {
+          // Throws this error in Firefox
+          setTimeout(() => {
```
robertu7, line 39: will this get into dead loop?

robertu7 (approved):

src/components/Hook/useColorThief.tsx
pitb2022: Not at the moment, but I think it's best to add max _getColor count

pitb2022: Done

PR #3911, opened 2023-10-17
Unit Test / Pt.2

robertu7:
Base branch: `feat/unit-test-wrapper`

* Add tests for `<Dialog>`, `<Dropdown>`, `<Menu>`, `<List>`, `<TableView>`;

### DIFF
--- a/src/components/Popper/index.tsx
+++ b/src/components/Popper/index.tsx
@@ -152,6 +152,7 @@ export const hidePopperOnClick = (instance: PopperInstance) => {
   box.addEventListener('click', (event: any) => {
     const target = event.target as HTMLElement
 
+    console.log('click....')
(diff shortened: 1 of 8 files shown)


src/components/Popper/index.tsx
```diff
@@ -152,6 +152,7 @@ export const hidePopperOnClick = (instance: PopperInstance) => {
   box.addEventListener('click', (event: any) => {
     const target = event.target as HTMLElement
 
+    console.log('click....')
```
pitb2022, line 155: 👀

pitb2022 (approved):

src/components/Popper/index.tsx
robertu7: removed in another PR

PR #3932, opened 2023-10-25
Add trackers for universal authentication process

zeckli:
Changes includes:
- Put `UNIVERSAL_AUTH_TRIGGER` back, and also add more
- Add tracker for authenticate engaging and leaving events

### DIFF
--- a/src/views/Auth/Universal/index.tsx
+++ b/src/views/Auth/Universal/index.tsx
@@ -3,7 +3,7 @@ import dynamic from 'next/dynamic'
 import { useContext, useEffect, useState } from 'react'
 
 import { PATHS } from '~/common/enums'
-import { WalletType } from '~/common/utils'
+import { analytics, getTrigger, WalletType } from '~/common/utils'
 import {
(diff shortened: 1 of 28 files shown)


src/views/Auth/Universal/index.tsx
```diff
@@ -83,6 +83,24 @@ const UniversalAuth = () => {
     setFirstRender(false)
   }, [])
 
+  useEffect(() => {
+    const track = (url: string) => {
+      const as = window?.history?.state?.as
+      if (url === as || url === PATHS.HOME) {
```
robertu7, line 89: a bit confused about this since `src/views/Auth/Universal/index.tsx` is only render on `/login` and `/signup`.

robertu7 (approved):

src/views/Auth/Universal/index.tsx
zeckli: Here is about leaving auth page, so it's emitting when user click logo or `history.back()`.

PR #3953, opened 2023-10-30
feat(upload): adjust gif support

gary02:
resolve https://github.com/thematters/developer-resource/issues/253

### DIFF
--- a/src/common/utils/url.ts
+++ b/src/common/utils/url.ts
@@ -33,6 +33,7 @@ interface ToSizedImageURLProps {
   width: ToSizedImageURLSize
   height?: ToSizedImageURLSize
   ext?: 'webp'
+  disableAnimation?: boolean
 }
 
 export const changeExt = ({ key, ext }: { key: string; ext?: 'webp' }) => {
@@ -52,6 +53,7 @@ export const toSizedImageURL = ({
   width,
   height,
   ext,
+  disableAnimation,
 }: ToSizedImageURLProps) => {
   const assetDomain = process.env.NEXT_PUBLIC_CF_IMAGE_URL
     ? `${process.env.NEXT_PUBLIC_CF_IMAGE_URL}`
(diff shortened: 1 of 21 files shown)


src/components/Cover/index.tsx
```diff
@@ -23,6 +23,7 @@ export const Cover: React.FC<React.PropsWithChildren<CoverProps>> = ({
         url={url}
         width={1376}
         disabled={isFallback || inEditor}
+        disableAnimation={true}
```
robertu7, line 26: this component is used as user profile cover, tag profile cover, etc. and it's supposed to support animated GIF.

src/components/TagDigest/Feed/index.tsx
```diff
@@ -126,6 +126,7 @@ const Feed = ({ tag, ...cardProps }: TagDigestFeedProps) => {
                   url={tag.cover || IMAGE_TAG_COVER.src}
                   width={144}
                   height={144}
+                  disableAnimation={true}
```
robertu7, line 129: please confirm if tag cover support animated GIF in feed.

src/components/TagDigest/Sidebar/index.tsx
```diff
@@ -65,6 +65,7 @@ const Sidebar = ({ tag, ...cardProps }: TagDigestSidebarProps) => {
                 url={tag.cover || IMAGE_TAG_COVER.src}
                 width={144}
                 height={144}
+                disableAnimation={true}
```
robertu7, line 68: please confirm if tag cover support animated GIF in sidebar.

robertu7 (changes requested):

src/common/utils/url.ts
```diff
@@ -76,10 +78,14 @@ export const toSizedImageURL = ({
   const hostnameless = url.replace(urlDomain, ``)
   const key = hostnameless.replace('/public', '')
   const extedUrl = changeExt({ key, ext })
-  const postfix = height
+  let postfix = height
     ? `w=${width},h=${height},fit=crop`
     : `w=${width},h=${width * 4},fit=scale-down`
 
+  if (disableAnimation) {
```
robertu7, line 85: can disable by default instead since most of the use cases don't support animated GIF?

robertu7 (changes requested):

src/common/utils/url.ts
gary02: I think animation enable by default is a little more safe then disable by default ?

robertu7 (approved):

PR #3996, opened 2023-11-10
Feat/revise user profile cover

pitb2022:


src/components/Spinner/index.tsx
```diff
@@ -22,3 +22,21 @@ export const Spinner = () => {
     </div>
   )
 }
+
+export const SpinnerWhite = () => {
```
robertu7, line 26: can modify `<Spinner>` and add `color` prop (default to current's greyLight).

robertu7 (approved):

src/components/Spinner/index.tsx
pitb2022: done

PR #4047, opened 2023-12-01
fix(Expandable): refactor component

pitb2022:


src/components/Truncate/index.tsx
```diff
@@ -0,0 +1,231 @@
+import {
```
robertu7, line 1: this component is hard to understand or maintain, better to provide more context or references

PR #4098, opened 2023-12-15
Feat/article normal state

pitb2022:
https://github.com/thematters/developer-resource/issues/329


src/components/DotDivider/styles.module.css
```diff
@@ -0,0 +1,5 @@
+.dot {
+  display: inline-flex;
+  justify-content: center;
+  width: 0.375px; /* 6px */
```
robertu7, line 4: rem

src/views/ArticleDetail/MetaInfo/index.tsx
```diff
@@ -21,33 +37,75 @@ const MetaInfo = ({
...
+  const { href } = toPath({ page: 'articleDetail', article })
 
   return (
     <section className={styles.info}>
+      <Media at="sm">
+        {/* TODO: Confirm display word length with product */}
+        <section className={styles.author}>
+          <UserDigest.Plain user={article.author} />
+          <section className={styles.dot}>
+            <DotDivider />
```
robertu7, line 53: can use CSS instead? (::before / ::after)

pitb2022: use ::before / ::after to fill space?

robertu7 (approved):

PR #4101, opened 2023-12-18
fix(strip-undefined): strip undefined value from referralCode

tx0c:
problem in `src/common/utils/oauth.ts` converted `undefined` to string `"undefined"`


robertu7 (approved):
Looks good. Any discussion about supporting more share channels with referral?

tx0c:
the undefined values didn't cause problems but statistics noises; need some tests

PR #4104, opened 2023-12-19
feat(node-v18): Bump deps: Node.js 18.x (client-side)

### LINKED ISSUE #3846: Bump deps: Node.js 18.x (client-side)

tx0c:
resolves #3846

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,9 +22,9 @@ jobs:
         uses: actions/checkout@master
 
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: '16.14'
+          node-version: '18'
           cache: 'npm'
 
       - name: Install Dependencies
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -22,9 +22,9 @@ jobs:
         uses: actions/checkout@master
 
       - name: Setup Node.js
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: '16.14'
+          node-version: '18'
```
robertu7, line 27: shall we fix it to a minor version, e.g. 18.19

robertu7: below same

robertu7 (approved):

.github/workflows/deploy.yml
tx0c: maybe; or may not be a big issue; the v18.19.x will not have another `v18.<next-minor>`

PR #4157, opened 2024-01-23
Feat/drawer component

pitb2022:
https://github.com/thematters/developer-resource/issues/335


src/views/ArticleDetail/styles.module.css
```diff
@@ -22,3 +22,16 @@
   padding: var(--spacing-loose) 0 0;
   margin: var(--spacing-loose) 0 var(--spacing-base);
 }
+
```
robertu7, line 25: unused?

src/components/Drawer/BaseDrawer.tsx
```diff
@@ -0,0 +1,100 @@
...
+        left: 0,
+        transform: 'translate3d(-100%, 0, 0)',
+        width: size,
+        height: '100vh',
+      }
+    case 'right':
+      return {
+        top: 0,
+        right: 0,
+        transform: 'translate3d(100%, 0, 0)',
```
robertu7, line 36: We should confirm whether the show/hide animation is smooth enough since there might be render-blocking issues caused by the children elements.

robertu7: and consider to add [`will-change`](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change) property for smoother transition.

robertu7 (approved):

PR #4193, opened 2024-02-20
Feat/comment jump

pitb2022:


src/views/ArticleDetail/Comments/CommentDetail/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+      let targetElement = comment
+      if (!!descendantId && comment) {
+        targetElement = comment.parentElement
+      }
+
+      if (!targetElement) {
+        return
+      }
+
+      targetElement.scrollIntoView({ behavior: 'smooth', block: 'start' })
```
robertu7, line 62: any differences from the current `jump.js`?

src/views/ArticleDetail/Comments/CommentDetail/index.tsx
```diff
@@ -0,0 +1,125 @@
...
+      const currentBgColor = targetElement.style.backgroundColor
+
+      setTimeout(() => {
```
robertu7, line 65: Can use pure CSS to implement instead? https://github.com/thematters/matters-web/blob/develop/src/components/Comment/Feed/styles.module.css#L2

lang/en.json
robertu7, line 1: image

src/components/Forms/CommentFormBeta/index.tsx
robertu7, line 1: the `@mention` should follow with a blank when reply a comment.

image

robertu7 (approved):

src/views/ArticleDetail/Comments/CommentDetail/index.tsx
pitb2022: jump.js will scroll the entire page, while we only need to scroll the comment drawer component

src/views/ArticleDetail/Comments/CommentDetail/index.tsx
pitb2022: The attributes that need to be changed are not the same for a first-level comment as for a second-level comment, and this way doesn't work

lang/en.json
pitb2022: Has fixed in next pr

PR #4203, opened 2024-02-28
Feat/Comment Misc

pitb2022:


src/components/CommentBeta/DropdownActions/CopyCommentButton.tsx
```diff
@@ -0,0 +1,29 @@
+import * as clipboard from 'clipboard-polyfill'
```
robertu7, line 1: as adding similar lib, we should refactor `<CopyToClipboard>` and remove `react-copy-to-clipboard` lib

robertu7 (approved):

PR #4207, opened 2024-03-04
refactor(components): migrate Head component from next/head to next-seo

### LINKED ISSUE #3838: Replace <Head> with garmeeh/next-seo

byhow:
referencing github issue #3838 and recommendation from next-seo docs that next-seo is better suited for this sort of things over next/head. most of the things here overlaps with app router's [metadata API](https://nextjs.org/docs/app/building-your-application/optimizing/metadata) too.

fix #3838

### DIFF
--- a/src/components/Head/index.tsx
+++ b/src/components/Head/index.tsx
@@ -66,8 +66,8 @@ export const Head: React.FC<HeadProps> = (props) => {
   }
 
   const getLocalLang = (props: HeadProps): LocalAlternativeLang[] => {
-    const alternativeLanguages: LocalAlternativeLang[] = []
     if (props.availableLanguages) {
+      const alternativeLanguages: LocalAlternativeLang[] = []
       for (const lang of props.availableLanguages) {
         // we only want these 3 languages so far
         if (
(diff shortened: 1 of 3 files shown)


src/components/Head/index.tsx
```diff
@@ -32,204 +34,206 @@ interface HeadProps {
...
+          : `matters,${process.env.NEXT_PUBLIC_SITE_DOMAIN},創作有價`,
+      },
+      {
+        name: 'viewport',
+        content:
+          'minimum-scale=1, initial-scale=1, width=device-width, shrink-to-fit=no, user-scalable=no, viewport-fit=cover',
+      },
+      {
+        name: 'monetization',
+        content: props.paymentPointer || '',
```
robertu7, line 136: Will `next-seo` render meta tags that `content` is nullish?

robertu7 (approved):
LGTM! 

Changed the based branch and will open another PR for Vercel preview.

src/components/Head/index.tsx
robertu7: can confirm empty tags are still be rendered.

image

byhow: Ah good catch! yeah this static config way of things isn't the best for ternary operations

byhow: temporary resort can be that we keep the ternary stuff still in `next/head` if we don't want to conditionally merging jsons all the time

robertu7: Can we use object spread with ternary?

```js
...props.paymentPointer ? { name: 'monetization', content: props.paymentPointer } : {},
...isProdServingCanonical ? {} : { name: 'googlebot', content: 'noindex, nofollow'},
```

PR #4208, opened 2024-03-04
fix(fallback-media-hash): as part of bulk redirects

tx0c:
for thematters/developer-resource#373


robertu7:
closed by https://github.com/thematters/matters-web/pull/4315

Issue #4196, opened 2024-02-21
Revise pathname of article detail page


robertu7:
### Checklist 
* Revise [pathname](https://github.com/thematters/matters-web/blob/develop/src/pages/%5Bname%5D/%5BmediaHash%5D.tsx) & [enums](https://github.com/thematters/matters-web/blob/develop/src/common/enums/route.ts) under `/a/[shortHash]`;
* Revise queries (by hash instead of mediaHash or articleId);
* Remove unused redirections (handled by edge https://github.com/thematters/developer-resource/issues/373);

---

### Story
- https://github.com/thematters/developer-resource/issues/272
- [Discussion](https://matterslab.slack.com/archives/C88CK7Q7L/p1709603507163969?thread_ts=1709589545.664879&cid=C88CK7Q7L)

PR #4209, opened 2024-03-05
feat(new-article-path): Revise pathname of article detail page #4196

tx0c:
partI for #4196;

remove the old scheme and switch from all listing pages (homepage, user_page, tag_listing, collection_listing) in separate PR,


tx0c:
original code in 82552000c ,  closed in #4211 instead

tx0c:
original code in 8255200 , closed in #4211 instead

PR #4210, opened 2024-03-05
Fix/archived comment detail

pitb2022:


src/views/ArticleDetail/Comments/CommentDetail/styles.module.css
```diff
@@ -0,0 +1,5 @@
+.container {
+  @media (--sm-up) {
+    width: 22.5rem;
```
robertu7, line 3: there are other places use this hardcoded value also, better to make it variable and reuse.

robertu7 (approved):

src/views/ArticleDetail/Comments/CommentDetail/styles.module.css
pitb2022: Will refactor in Donation drawer PR

PR #4227, opened 2024-03-12
feat(new-article-path): Revise pathname of article detail page #4196

tx0c:
for #4196


src/pages/a/[shortHash].tsx
```diff
@@ -0,0 +1,3 @@
+import ArticleDetail from '~/views/ArticleDetail'
```
Comment (robertu7) [inline src/pages/a/[shortHash].tsx:1]: did we confirm to use `/a` instead of `/p`?

PR #4275, opened 2024-04-15
Fix/article comment bug

pitb2022:
https://www.notion.so/matterslab/Zhao-Nan-9bffdb932e32435bbd799ae742886c9e

### DIFF
--- a/src/components/Dialogs/CommentFormBetaDialog/CommentForm/index.tsx
+++ b/src/components/Dialogs/CommentFormBetaDialog/CommentForm/index.tsx
@@ -68,6 +68,7 @@ const CommentForm: React.FC<CommentFormProps> = ({
   }:${replyToId || 0}`
   const formId = `comment-form-${commentDraftId}`
   const formRef = useRef<HTMLFormElement>(null)
+  // TODO: beware of the breaking change by the new article URL
   const { getQuery, router, routerLang } = useRoute()
   const mediaHash = getQuery('mediaHash')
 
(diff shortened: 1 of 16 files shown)


src/components/Dialogs/CommentFormBetaDialog/CommentForm/index.tsx
```diff
@@ -61,6 +68,8 @@ const CommentForm: React.FC<CommentFormProps> = ({
   }:${replyToId || 0}`
   const formId = `comment-form-${commentDraftId}`
   const formRef = useRef<HTMLFormElement>(null)
+  const { getQuery, router, routerLang } = useRoute()
+  const mediaHash = getQuery('mediaHash')
```
robertu7, line 72: beware of the breaking change by the new article URL

src/components/Dialogs/CommentFormBetaDialog/CommentForm/styles.module.css
```diff
@@ -10,6 +10,18 @@
...
   overflow-y: auto;
+
+  & > div {
+    height: 100%;
+
+    & > div {
+      height: 100%;
+
+      & > div {
+        height: 100%;
```
robertu7, line 21: it's not a good practice to select elements with nested selectors, is there any other solution (e.g. class name)?

robertu7 (approved):

src/components/Dialogs/CommentFormBetaDialog/CommentForm/index.tsx
pitb2022: TODO note has been added

src/components/Dialogs/CommentFormBetaDialog/CommentForm/styles.module.css
pitb2022: Use global classname

PR #4301, opened 2024-04-26
Confirming whether to leave the page

pitb2022:


src/components/Context/CommentDrafts/index.tsx
```diff
@@ -0,0 +1,134 @@
+import { useApolloClient } from '@apollo/react-hooks'
+import { createContext, ReactNode, useEffect, useState } from 'react'
+import { useIntl } from 'react-intl'
+
+import COMMENT_DRAFT from '~/components/GQL/queries/commentDraft'
+import { useRoute } from '~/components/Hook'
+
+type draft = string
```
robertu7, line 8: `type Draft`? or maybe just use `string` in L13-L14

robertu7 (approved):

PR #4330, opened 2024-05-07
feat(article): add page to handle legacy article url redirection

robertu7:
refs:
* https://github.com/thematters/developer-resource/issues/373
* https://github.com/thematters/matters-web/issues/4196
* https://github.com/thematters/matters-web/pull/4315

### DIFF
--- a/src/pages/[name]/[idOrMediaHash]/index.tsx
+++ b/src/pages/[name]/[idOrMediaHash]/index.tsx
@@ -1,0 +1,139 @@
+import { useQuery } from '@apollo/react-hooks'
+import gql from 'graphql-tag'
+import { useRouter } from 'next/router'
+import { useEffect } from 'react'
+
+import { toGlobalId, toPath } from '~/common/utils'
+import { EmptyLayout, SpinnerBlock, Throw404, useRoute } from '~/components'
+import {
+  GetShortHashByIdQuery,
+  GetShortHashByMediaHashQuery,
+} from '~/gql/graphql'
+
+/**
+ * FIXME: can retire this page when Cloudflare Bulk Redirects is ready
+ */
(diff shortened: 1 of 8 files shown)


src/pages/[name]/[idOrMediaHash]/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    { variables: { id: toGlobalId({ type: 'Article', id: articleId }) } }
+  )
+  const shortHash =
+    data?.node?.__typename === 'Article' ? data.node.shortHash : ''
+
+  useEffect(() => {
+    if (!shortHash) return
+
+    const path = toPath({ page: 'articleDetail', article: { shortHash } })
+    router.push(path.href)
```
Comment (tx0c) [inline src/pages/[name]/[idOrMediaHash]/index.tsx:52]: is it better to call `router.replace` ?

https://nextjs.org/docs/pages/api-reference/functions/use-router#routerreplace

src/pages/[name]/[idOrMediaHash]/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+    GET_SHORT_HASH_BY_MEDIA_HASH,
+    { variables: { mediaHash } }
+  )
+  const shortHash = data?.article?.shortHash
+
+  useEffect(() => {
+    if (!shortHash) return
+
+    const path = toPath({ page: 'articleDetail', article: { shortHash } })
+    router.push(path.href)
```
Comment (tx0c) [inline src/pages/[name]/[idOrMediaHash]/index.tsx:87]: and here,

src/pages/[name]/[idOrMediaHash]/index.tsx
Comment (robertu7) [reply src/pages/[name]/[idOrMediaHash]/index.tsx]: nice catch, resolved

PR #4335, opened 2024-05-09
Fix/product-pre-test-part-I

pitb2022:


src/common/styles/utils/content.article.css
```diff
@@ -58,8 +56,7 @@
    ========================================================================== */
   & figure {
     position: relative;
-    padding: var(--ar17-200) 0;
-    margin: 0;
+    margin: var(--ar17-200) 0;
```
robertu7, line 59: Cannot simply change padding to margin, need to consider margin collapse here. 

For instance, there is margin collapse on two adjacent sibling `figure`, the spacing is `--ar17-200` (should be `--ar17-200` * 2)

src/views/ArticleDetail/Toolbar/FloatToolbar/index.tsx
```diff
@@ -123,7 +129,18 @@ const FloatToolbar = ({
...
+              if (!viewer.isAuthed) {
+                window.dispatchEvent(
+                  new CustomEvent(OPEN_UNIVERSAL_AUTH_DIALOG, {
+                    detail: {
+                      trigger: UNIVERSAL_AUTH_TRIGGER.support,
+                    },
+                  })
+                )
+                return
+              }
```
robertu7, line 142: else ... `toggleDonationDrawer`

robertu7 (approved):

PR #4344, opened 2024-05-09
Fix/product pre test part ii

pitb2022:


src/views/ArticleDetail/Support/SupportAuthor/Tabs/index.tsx
```diff
@@ -11,7 +12,9 @@ type Props = {
 }
 
 const DonationTabs = ({ recipient, currency, setCurrency }: Props) => {
+  const viewer = useContext(ViewerContext)
   const creatorAddress = recipient.info.ethAddress
+  const hasLikeId = viewer.liker.likerId !== null
```
robertu7, line 17: `viewer.liker.likerId` might also be `undefined`, can use `!!viewer.liker.likerId` instead?

image

robertu7 (approved):

PR #4358, opened 2024-05-13
Fix/production test

pitb2022:


src/components/Spacer/index.tsx
```diff
@@ -2,6 +2,7 @@ import styles from './styles.module.css'
 
 interface SpacerProps {
   size?:
+    | 'xxtight'
```
robertu7, line 5: should refactor to new convention: `sp*`

robertu7 (approved):

src/components/Spacer/index.tsx
pitb2022: add TODO

Issue #4251, opened 2024-03-26
Allow pasting in Mattes ID setup


zeckli:
### References
- [thread](https://matterslab.slack.com/archives/GA0R09FHN/p1709654061745509)

PR #4362, opened 2024-05-13
feat(inputstep): enable paste in matters id input

byhow:
In the previous state, pasting Matters ID during registration was disabled to alert user the importance of this UUID. However, this presents UX drawbacks to users for unexpected behavior. Reverting to default behavior.

re #4251


byhow:
e2e test timeout due to cloudflare security setting

robertu7 (approved):

robertu7:
First PR 🚀

Issue #4250, opened 2024-03-26
Universal date time format


zeckli:
### References
- [doc](https://www.notion.so/matterslab/89907b23ec664c51b96f91fea72b60d4)

PR #4385, opened 2024-05-14
feat(datetime-format): unify date time format across matters platform

byhow:
Adds date time format standard for matters platform including truncated time format for `Feed` feature that is coming in the near future. This universal time format aims to keep the standard in sync instead of having absolute vs relative time formats

BREAKING CHANGE: absolute time option is removed from date time formatting

fix [#4250](https://github.com/thematters/matters-web/issues/4250)


src/common/utils/datetime/absolute.test.ts
robertu7, line 1: we can add them back if needed, feel free to remove absolute format.

robertu7 (approved):
LGTM

src/common/utils/datetime/absolute.test.ts
byhow: Got it, absolute format was used for a fallback here and also just in case if we found this being useful in the future, but maybe we can get rid of it sometimes too

Issue #4368, opened 2024-05-14
Add logo button to the top of article page


zeckli:
### Description
On mobile device, there's no simple redirect to the home when viewing article pages. Therefore, we need to put a back home button.

### References
- [report thread PR-066](https://www.notion.so/matterslab/Logo-fc264b60a3be4df59d71f7ad34e7841b?pvs=4)
- [design](https://www.figma.com/design/gs3w7fKdHqk9u1nnDxPYlZ/%E4%BD%9C%E5%93%81%E9%A0%81?node-id=976%3A39468&t=DZjkCV0WeUil9aTp-1)

---
### Story
- https://github.com/thematters/developer-resource/issues/372

PR #4395, opened 2024-05-16
feat(articledetail): add matters logo as back button when display is mobile

byhow:
A preview of the logo on mobile display:
image

fix: #4368


public/static/icons/logo-mobile.svg
robertu7, line 1: looks like the same icon as `public/static/icons/logo.svg` except the color? 

can change `fill="#333"` to `fill: currentColor` so that we can specify the color in `<Icon>`.

src/views/ArticleDetail/index.tsx
```diff
@@ -340,6 +341,16 @@ const BaseArticleDetail = ({
       </Media>
 
       <section className={styles.content}>
+        <Media
+          at="sm"
+          style={{
```
robertu7, line 346: better to avoid inline styles

robertu7 (changes requested):

byhow:
comments addressed. new logo now looks like this: 
image

public/static/icons/logo-mobile.svg
byhow: nice, thanks for the pointer to the documentation 🙌

src/views/ArticleDetail/index.tsx
byhow: refactored

robertu7 (approved):
LGTM!

PR #4411, opened 2024-05-20
feat(memory-metrics): add aws-nextjs-metrics for memory reporting

tx0c:


.ebextensions/eb-memory-monitor.config
```diff
@@ -1,19 +1,16 @@
+files:
+  "/home/ec2-user/aws-nextjs-metrics/aws-nextjs-metrics.mjs" :
+    mode: "000755"
+    owner: 1000
+    group: 100
+    source: https://raw.githubusercontent.com/thematters/matters-web/3eb0d35533baa8720d3f92610610def0059fd0b2/bin/aws-nextjs-metrics.mjs
```
robertu7, line 6: this is not a reliable source

robertu7 (approved):

.ebextensions/eb-memory-monitor.config
tx0c: the source needs to be changed to develop or main branch later;

PR #4428, opened 2024-05-21
feat(memory-metrics): add aws-nextjs-metrics for memory reporting

tx0c:
the source to be changed to 'develop' branch later;

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "test:unit": "vitest run",
     "test:unit:coverage": "vitest run --coverage",
     "test:unit:report": "vitest --ui --coverage.enabled=true",
-    "start": "PORT=\"${PORT:-3000}\"; cross-env NODE_OPTIONS='--no-experimental-fetch' next start -p $PORT",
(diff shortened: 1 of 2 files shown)


robertu7 (approved):

package.json
```diff
@@ -16,7 +16,7 @@
     "test:unit": "vitest run",
     "test:unit:coverage": "vitest run --coverage",
     "test:unit:report": "vitest --ui --coverage.enabled=true",
-    "start": "PORT=\"${PORT:-3000}\"; cross-env NODE_OPTIONS='--no-experimental-fetch' next start -p $PORT",
+    "start": "PORT=\"${PORT:-3000}\"; cross-env NODE_OPTIONS='--no-experimental-fetch --max-old-space-size=$[3000]' next start -p $PORT",
```
robertu7, line 19: the instance of dev env should fixed to at least 4GB memory if we hardcoded here.

tx0c: yes, it's still the problem, no easy way to set a `Math.max(MemTotal/2, MemTotal-1GB)` number here,

PR #4451, opened 2024-05-24
Fix/remove has payment password context

pitb2022:


robertu7 (changes requested):
can be fixed by just reverting the commit df6fc7a1546f8ea30ca433076cb96440a1963d7d 

related issue: https://github.com/thematters/matters-web/pull/4443

robertu7 (approved):

Issue #4370, opened 2024-05-14
Adjust page down


zeckli:
### Description
Adjust page down behavior

### References
- [report thread PR-040](https://www.notion.so/matterslab/Page-Down-da76e985185f41e7a4d61e015df284d8?pvs=4)

---
### Story
- https://github.com/thematters/developer-resource/issues/372

PR #4491, opened 2024-05-28
fix(article-detail): adjust page up/down scroll height for floating toolbar

byhow:
Since the presence of a floating toolbar, the normal scroll height for page up/down (fn + up/down key on OSX) hides some pixels. Now it will offset the height by the height of the toolbar

close: #4370

### DIFF
--- a/src/views/ArticleDetail/index.tsx
+++ b/src/views/ArticleDetail/index.tsx
@@ -32,6 +32,7 @@ import {
   useEventListener,
   useFeatures,
   useIntersectionObserver,
+  useNativeEventListener,
   usePublicQuery,
   useRoute,
   ViewerContext,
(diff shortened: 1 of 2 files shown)


src/views/ArticleDetail/index.tsx
```diff
@@ -539,6 +540,28 @@ const ArticleDetail = ({
...
+    if (
+      event.code.toLowerCase() === 'pagedown' ||
+      event.code.toLowerCase() === 'pageup' ||
+      event.code.toLowerCase() === 'space'
+    ) {
+      event.preventDefault()
+      const remInPixels = parseFloat(
+        getComputedStyle(document.documentElement).fontSize
+      )
+      const scrollDirection = event.code.toLowerCase() === 'pagedown' ? 1 : -1
```
robertu7, line 555: `space` should be `1`

src/views/ArticleDetail/index.tsx
```diff
@@ -539,6 +540,28 @@ const ArticleDetail = ({
     loadPrivate()
   }, [article?.shortHash, viewer.id])
 
+  // floating toolbar is blocking some text when pressing page up and down
+  // offsetting it with the height of floating toolbar
+  useNativeEventListener('keydown', (event: KeyboardEvent) => {
+    if (
+      event.code.toLowerCase() === 'pagedown' ||
```
robertu7, line 547: there is a key code mapping at `src/common/enums/keyValue.ts`

robertu7 (changes requested):

src/views/ArticleDetail/index.tsx
byhow: good call! just updated along with the `keyValue` enum

robertu7 (approved):

Issue #4486, opened 2024-05-28
Moment - Follow Page Tab Switch


byhow:
As a user, I want to be able to switch tabs views from `all` items and `article only` in the new main page.

[figma](https://www.figma.com/design/0d2OccTXpQVsYtY2fABX4X/%E5%8B%95%E6%85%8B?node-id=2265-59212&t=mWfZwxwwD8IvZu3t-0)

PR #4544, opened 2024-06-12
feat(layout): add popover for new journal entry

byhow:
I was working on the `Button` stories prior, so I happen to refactor that page as well. Others are in the `Tabs` story.

re: #4486


src/stories/components/Tabs/JournalTabsGroup.tsx
```diff
@@ -0,0 +1,54 @@
+import classNames from 'classnames'
+import React, { useState } from 'react'
+import { useIntl } from 'react-intl'
+
+import styles from '../../../views/ArticleDetail/AuthorSidebar/Tabs/styles.module.css'
```
robertu7, line 5: move `~/views/ArticleDetail/AuthorSidebar/Tabs` to `~/components` and make it reusable?

robertu7 (approved):

src/stories/components/Tabs/JournalTabsGroup.tsx
byhow: The logic in `~/views/ArticleDetail/AuthorSidebar/Tabs` is pretty coupled actually, see https://github.com/thematters/matters-web/blob/25c68f503654a379071ea09619096d69bdc41985/src/views/ArticleDetail/AuthorSidebar/Tabs/index.tsx#L46

Maybe in the future when we figured those conditions can be refactored out of the component we can make the move. I can add a TODO for now

robertu7: should abstract those UI elements and make it reusable without specific business logic.

robertu7: > I can add a TODO for now
looks good for now. we can do it when integrating it into following page.

PR #4554, opened 2024-06-12
feat(editor): use keyboard up/down to select an mention option

robertu7:


src/components/Editor/Article/extensions/mention/suggestion.ts
```diff
@@ -68,7 +76,7 @@ export const makeMentionSuggestion = ({
         }
 
         popup = tippy('body', {
-          getReferenceClientRect: props.clientRect,
+          getReferenceClientRect: props.clientRect as GetReferenceClientRect,
```
byhow, line 79: not the biggest fan of casting but also i dont have much context working with Tippy

byhow (approved):
just nits

Issue #4607, opened 2024-06-25
Update article card in icymi feed


zeckli:
### Description
- Remove timestamp from article card
- Remove meta data from article card footer

### References
- [product doc](https://www.notion.so/matterslab/124280f0955c4d1cbf48ad8220a2f3c9?pvs=4)

PR #4614, opened 2024-07-01
refactor(article-digest): remove metadata for featured

byhow:
fix https://github.com/thematters/matters-web/issues/4607

### DIFF
--- a/src/views/Home/Feed/IcymiCuratedFeed/index.tsx
+++ b/src/views/Home/Feed/IcymiCuratedFeed/index.tsx
@@ -115,6 +115,7 @@ export const IcymiCuratedFeed = ({ recommendation }: IcymiCuratedFeed) => {
             <List.Item key={article.id}>
               <ArticleDigestFeed
                 article={article}
+                includesMetaData={false}
                 onClick={() =>
                   onClickArticle('article', cardArticleNum + i, article.id)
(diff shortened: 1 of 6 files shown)


src/views/Home/Feed/IcymiCuratedFeed/index.tsx
```diff
@@ -115,6 +115,7 @@ export const IcymiCuratedFeed = ({ recommendation }: IcymiCuratedFeed) => {
             <List.Item key={article.id}>
               <ArticleDigestFeed
                 article={article}
+                includesMetaData={false}
```
robertu7, line 118: `<IcymiCuratedFeed>` is a optional section of ICYMI feed, we should also disable `includesMetaData` for [infinite list for ICYMI feed](https://github.com/thematters/matters-web/blob/develop/src/views/Home/Feed/MainFeed/index.tsx#L228).

robertu7 (changes requested):

src/views/Home/Feed/IcymiCuratedFeed/index.tsx
byhow: cool, thx for the pointer! updated in the latest commit

robertu7 (approved):

PR #4618, opened 2024-07-03
build: remove extraneous dev deps

byhow:
- removed dev deps that are unused
- parallelize dev npm script

re: [#4640](https://github.com/thematters/matters-web/issues/4640)


package.json
```diff
@@ -175,7 +172,6 @@
     "eslint-plugin-simple-import-sort": "^10.0.0",
     "eslint-plugin-storybook": "^0.6.15",
     "eslint-plugin-testing-library": "^6.1.0",
-    "file-loader": "^6.2.0",
```
byhow, line 178: still need this

PR #4644, opened 2024-07-11
feat(MomentDigest): Image Lightbox

pitb2022:


src/components/MomentDigest/Assets/index.tsx
```diff
@@ -17,19 +26,102 @@ const fragments = {
...
-      ))}
+      <Gallery options={{}}>
+        {loadedAssets.map((asset) => (
+          <Item
+            key={asset.id}
+            original={asset.path}
+            thumbnail={toSizedImageURL({
+              url: asset.path,
+              width: width,
+              disableAnimation: true,
```
robertu7, line 106: make it same props as L119 so it can load from cache instead of new request, e.g. remove `disableAnimation` or add `disableAnimation` to L119

src/components/MomentDigest/Assets/index.tsx
```diff
@@ -17,19 +26,102 @@ const fragments = {
...
+  }, [assets])
+
+  if (!loadedAssets) {
+    return (
+      <section
+        className={styles.assets}
+        data-test-id={TEST_ID.MOMENT_DIGEST_ASSETS}
+      >
+        {assets.map((asset) => (
+          <ImagePlaceholder key={asset.id} />
```
robertu7, line 81: can use default image as fallback if any (request to designer?)

ref: https://github.com/thematters/matters-web/blob/4b3529f40f1564b9c3ea88062db7aa5c174f568b/src/components/ArticleDigest/Curated/index.tsx#L75

robertu7 (approved):

src/components/MomentDigest/Assets/index.tsx
pitb2022: https://www.figma.com/design/0d2OccTXpQVsYtY2fABX4X/%E5%8B%95%E6%85%8B?node-id=1960-72692&m=dev

PR #4662, opened 2024-07-20
fix(campaign): fix missing click handler of apply button; new tab to open event link;

robertu7:


src/views/CampaignDetail/InfoHeader/index.tsx
```diff
@@ -72,20 +71,22 @@ const InfoHeader = ({ campaign }: InfoHeaderProps) => {
...
-                    placement="left"
-                  >
-                    <FormattedMessage
-                      defaultMessage="Event Information"
-                      id="buf5vO"
-                    />
-                  </TextIcon>
-                </a>
-              </Link>
+              <a
```
byhow, line 74: for future reference, `<Link>` does not renders `<a>` tags internally till Next 13, is this why this is reverted to a Link?

byhow (approved):

Issue #4625, opened 2024-07-08
Notices of collection like feature


zeckli:
### Description
To increase engagement between collection author and reader, we'll introduce collection like feature:

- [ ] New notice for collection like feature
- [ ] Update notice page for collection like

### References
- [design](https://www.figma.com/design/DDOV572OhW5hpEJsEPk9tE/%E6%94%B9%E9%80%B2?node-id=238-608&t=U4g1wmbztKG4YGaD-0)

PR #4686, opened 2024-07-26
feat(notice): collection like

byhow:
re: #4625


src/common/utils/text/collection.ts
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Truncates a title to a specified maximum length, while preserving tagged users.
+ *
+ * @param title - The title to truncate.
+ * @param maxLength - The maximum length of the truncated title.
+ * @param locale - The locale to determine the truncation rules. Defaults to 'en'.
+ * @returns The truncated title with preserved tagged users.
+ */
+export const truncateTitle = (
```
robertu7, line 11: looks similar to ~/utils/text/moment.ts. Can we merge it reusable?

* move to `~/utils/text/notice.ts`;
* rename to `truncateNoticeTitle`;
* add optional param to handle `@` for moment notice title;

robertu7 (changes requested):

src/common/utils/text/collection.ts
byhow: Got it, makes sense, will cherry pick the new commit to `feat/journal` when merged

robertu7 (approved):

PR #4795, opened 2024-08-28
Fix docker build

lizhineng:
I discovered that the _Dockerfile_ was out of date while setting up the project. This pull request updates the _Node_ version from _12_ to _18_ to match the requirements defined in _package.json_.

Additionally, the inclusion of the _.npmrc_ file ensures that we have the same _npm_ behavior inside Docker as we would on the host system. I believe this is due to the `legacy-peer-deps` setting.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:12.16
+FROM node:18
 
 # install os level packages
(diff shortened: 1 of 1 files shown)


lizhineng:
I'll try resolving peer dependency issues on another PR.

robertu7 (approved):
Looks good!

PR #4829, opened 2024-09-02
Drops --legacy-peer-deps

lizhineng:
## Motivation

The use of `--legacy-peer-deps` can often mask underlying issues in a project's dependency graph. By continuing to use this flag, we might inadvertently ignore critical warnings or errors related to peer dependencies. Dropping the flag encourages a more stable and maintainable project in the long term.

## Apollo Client

Given that many of the packages associated with Apollo Client _v2_ have been discontinued, I’ve decided to upgrade to _v3_ to ensure continued support and access to the latest features.

The following changes have been implemented following the guidance provided in the official [migration guide](https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration):

- Remove `@apollo/react-common`
- Update the importing from `@apollo/react-hooks` to `@apollo/client`
- Update the importing from `@apollo/react-ssr` to `@apollo/client/react/ssr`
- Remove `apollo-cache-inmemory` package
- Update the importing from `apollo-cache` to `@apollo/client/cache`
- Update the importing from `apollo-client` to `@apollo/client`
- Update the importing from `apollo-link` to `@apollo/client`
- Update the importing from `apollo-link-context` to `@apollo/client/link/context`
- Update the importing from `apollo-link-error` to `@apollo/client/link/error`
- Remove `apollo-link-persisted-queries` package
- Update the importing from `apollo-utilities` to `@apollo/client/utilities`
- Update the importing from `@apollo/react-testing` to `@apollo/client/testing`

### InMemoryCache

In Apollo Client _v3_, objects without unique identifiers are not merged by default. In our case, this applies to the _Official_, _Recommendation_, and _UserInfo_ objects. I’ve configured the store to merge these objects regardless, but please keep a close eye on this data. I’m not certain if this change could potentially disrupt the business logic.

The `IntrospectionFragmentMatcher` API has been removed and replaced by the new `possibleTypes` option. To align with this change, the artifact at `src/common/gql/fragmentTypes.json` and the build script at `bin/buildFragmentTypes.js` have been updated to accommodate the new data structure. For more details, please refer to the documentation on [Defining `possibleTypes` manually](https://www.apollographql.com/docs/react/data/fragments/#defining-possibletypes-manually).

The `client. writeData` has been removed and replaced by `client.cache.modify`.

### Persisted Queries

[...]

PR #4840, opened 2024-09-03
Drops --legacy-peer-deps from #4829

robertu7:


robertu7:
@lizhineng you can try to preview here: https://matters-web-git-fix-peer-dependency-matters.vercel.app/

I found there are some issue on local cache of article page. You can reproduce:
1. Open the homepage;
2. Click on any article;
3. Click on any link in the article page;
4. Go back to article page: expected not to show the loading state;

lizhineng:
The article refresh you’re seeing is actually expected. The `useEffect` hook runs whenever the component gets mounted back to the page.

In Apollo Client _v2_, the loading state didn’t visibly update, but the HTTP requests were still happening behind the scenes. I’ve recorded a video showing how this works in production for reference.

https://github.com/user-attachments/assets/3ee27e1b-3a9b-4575-a34e-7af25965a9ee

Here's a [PR](https://github.com/thematters/matters-web/pull/4851) that applies a quick fix for the "issue." I don’t have a complete understanding of the business logic, so feel free to close this if it doesn't work as expected or causes any problems.

robertu7:
Thanks for your debugging, i think this bug it's brought a fix from https://github.com/thematters/matters-web/pull/3904, however, it's not useful anymore. 

> I don’t have a complete understanding of the business logic, so feel free to close this if it doesn't work as expected or causes any problems.
We still need the `useEffect` hook to fetch the "private data" to ensure that the data and rendering are correct for the current user and article state. We will close your PR and push the commit later.

lizhineng:
> image

Is this error caused by these PR changes? How do you reproduce it and maybe I can look it up.

PR #4873, opened 2024-09-13
refactor: parse URL with URI.js for better compatibility and maintainability 

robertu7:


src/pages/_document.tsx
```diff
@@ -57,6 +57,13 @@ class MattersDocument extends Document<MattersDocumentProps> {
             strategy="lazyOnload"
             crossOrigin="anonymous"
           />
+
+          <Script
+            async
+            src="https://cdn.jsdelivr.net/npm/uricli@latest/index.js"
```
guoliu, line 63: I believe we can use [browser API](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) instead of an external package. Also, if we need to introduce external packages, please use npm instead.

src/common/enums/csp.ts
```diff
@@ -37,6 +37,9 @@ const SCRIPT_SRC = [
 
   // Stripe
   'js.stripe.com',
+
+  // JSDelivr
+  'cdn.jsdelivr.net',
```
guoliu, line 42: We should avoid adding new items to this whitelist as much as possible.

guoliu (changes requested):

PR #4886, opened 2024-10-07
feat(copy): revise "follow" copies

robertu7:


src/views/About/Footer/index.tsx
```diff
@@ -90,8 +90,8 @@ const Footer = () => {
             <section className={styles.followUs}>
               <h2>
                 <Translate
-                  zh_hant="關注我們"
-                  zh_hans="關注我們"
+                  zh_hant="追蹤我們"
+                  zh_hans="追蹤我們"
```
zeckli, line 94: Here should be 關注 ?

zeckli (approved):

